PR #71, opened 2014-09-10
Enable head tag attributes via config

net02:
The 'head' attributes set via config were used as html attributes


rande:
I guess this is a BC break. Cannot we do something to preserve the old behavior ?

rande:
This PR need to be rebase .

net02:
Rebase done.
Looking back at commit history I understood how we got to this possible BC stage:
* SeoPage had only *html_attributes*, configured through the 'head' config parameter
* then *head_attributes* were added to SeoPage, without any config integration
* now we're stuck with a need for a BC or some counter-intuitive config parameter naming

I think we should divert from the current naming, and preserve the old 'head to html attributes' with a deprecation warning maybe?

net02:
Here's my take

core23:
Please rebase and remove the unrelated commits @net02

net02:
@core23 there it is, rebased and cleaned up!

DependencyInjection/Configuration.php
```diff
@@ -29,6 +29,16 @@ public function getConfigTreeBuilder()
                             ->useAttributeAsKey('attribute')
                             ->prototype('scalar')->end()
                         ->end()
+                        ->arrayNode('html_tag')
```
core23: Can't we use the the keys **html** and **head** for this?

net02: It used to before @rande pointed out it would be a BC break. Sadly previous commits are gone with the squash so you can't see this PR history.

The problem is the current **head** key [actually configures the <html> attributes](https://github.com/sonata-project/SonataSeoBundle/blob/master/DependencyInjection/SonataSeoExtension.php#L58). To keep BC **head** and **html_tag** config are merged (raising a deprecation warning on **head**). [Here's](#issuecomment-84668910) more on the reasoning behind it.

core23:
I would merge this as it is. Can you create a new PR for the next major which changes the keys back to ``head`` and ``html``.

core23:
@OskarStark @greg0ire Can you give a second review before merging?

Tests/DependencyInjection/ConfigurationTest.php
```diff
@@ -47,8 +49,8 @@ public function testKeysAreNotNormalized()
     {
         $values = array(
             'page' => array(
-                'head'  => array('data-example' => 'abc-123'),
-                'metas' => array(
+                'html_tag' => array('data-example' => 'abc-123'),
```
OskarStark: ``head_tag`` missing in test?

DependencyInjection/SonataSeoExtension.php
```diff
@@ -53,9 +53,14 @@ protected function configureSeoPage(array $config, ContainerBuilder $container)
     {
         $definition = $container->getDefinition($config['default']);
 
+        if (count($config['head'])) {
```
OskarStark: this should be done in ``Configuration.php``

Resources/doc/reference/installation.rst
```diff
@@ -55,7 +55,7 @@ file.
                 charset:
                     UTF-8:    ''
 
-            head:
+            html_tag:
```
OskarStark: looks like installation.rst and usage.rst does not say sth about the ``head_tag``

DependencyInjection/Configuration.php
greg0ire: Reading this piece of code, I have no idea what all this is about. Can you add a call to `info()` on each option, so that it is documented?

OskarStark:
ping @core23

core23:
Ping @net02

Tests/DependencyInjection/ConfigurationTest.php
```diff
@@ -47,8 +49,8 @@ public function testKeysAreNotNormalized()
     {
         $values = array(
             'page' => array(
-                'head'  => array('data-example' => 'abc-123'),
```
Soullivaneuh: Legacy `head` section should be tested and marked as as legacy.

Soullivaneuh:
According to the new [Sonata version management](https://github.com/sonata-project/SonataAdminBundle/issues/3053) and next major [release plan](https://github.com/sonata-project/SonataAdminBundle/issues/3731), this project has been refactored regarding branching and versioning.

If you see this message, your PR concerns a patch or a minor release and is not targeting the right branch.

So I'm closing this one, but don't see it as a refusal. If you think your work is still relevant and want to continue, feel free to reopen it on the right branch (e.g. the default one).

Regards.

Issue #86, opened 2015-03-17
Unrecognized option "default" under "sonata_seo"


rasanga:
After upgrading "sonata-project/seo-bundle" to version 1.1.10 from 1.1.8  I am getting following error message:

[Symfony\Component\Config\Definition\Exception\InvalidConfigurationException]  
  Unrecognized option "default" under "sonata_seo"                               
                                                     
I checked the CHANGEDLOG but I could not find anything regards but seems to be the configuration has been changed, my sonata_seo configuration looking like following:

# app/config/config.yml
sonata_seo:
    default:          sonata.seo.page.default
    encoding:         UTF-8
    page:
        title:            Sonata Project
        metas:
            name:
                keywords:             foo bar
                description:          The description
                robots:               index, follow

nlzet:
I had the same problem, I think I copied my setup from the sandbox. From what i've found quickly the option "default" belongs to your "page" option and not as a main option.

So replace:

    sonata_seo:
        default: sonata.seo.page.default
        encoding: UTF-8
        page:
            title: Sonata Project
            ...

With:

    sonata_seo:
        encoding: UTF-8
        page:
            default: sonata.seo.page.default
            title: Sonata Project
            ...

OskarStark:
:+1: thx @mvhirsch

nicolasricci:
documentation update in progress #90

OskarStark:
ping @rande why is the documentation still the old one on the website?

can you please update this?
https://sonata-project.org/bundles/seo/master/doc/reference/installation.html

rande:
@OskarStark the documentation is now uptodate

OskarStark:
thank you!

2015-09-02 18:38 GMT+02:00 Thomas <notifications@github.com>:

didando8a:
Perfect!!! Thanks

PR #87, opened 2015-03-23
re-arrange default key under page. refs #86

mvhirsch:
Had the same issue as described in #86.
I followed the installation guide, so I think this should fix the problem for others.

### DIFF
--- a/Resources/doc/reference/installation.rst
+++ b/Resources/doc/reference/installation.rst
@@ -27,10 +27,10 @@ file.
 
     # app/config/config.yml
     sonata_seo:
-        default:          sonata.seo.page.default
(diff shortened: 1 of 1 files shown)


OskarStark:
:+1: thx @mvhirsch

mvhirsch:
Why do the builds fail? It was just a documentation update ...
Can anyone please help me with this?

OskarStark:
this error comes from the sonata core bundle i think

@rande @Soullivaneuh can you please check that

anyway, we should update this bundle to make it compatible with the current SonataCoreBundle

anyway, please merge this PR

Soullivaneuh:
Thank you @mvhirsch.

PR #91, opened 2015-04-09
Fix twitter sdk twig block name

nlzet:
block name was wrong in twig template. It worked like it should but you cannot override the block correctly.


nlzet:
ping @rande , can this be merged?

PR #94, opened 2015-06-03
Fixed namespace in breadcrumb docs

adamquaile:


adamquaile:
The documentation here doesn't match the actual code, so it's fairly pointless to merge this typo fix. I may submit another PR if/when I figure out how it actually works.

PR #102, opened 2015-10-11
Allow Twig 2

WouterJ:


Soullivaneuh:
Twig is not stable yet. Is this version at least feature freezed?

WouterJ:
@Soullivaneuh no, it isn't. But it is always possible to support at least the latest Twig 1.x version and Twig 2.0, so I think we should already allow installing the bundle with Twig 2.

Soullivaneuh:
I would prefer a stable version.

You can maybe force 2.x on Travis to check if its work while waiting this.

Soullivaneuh:
According to the new [Sonata version management](https://github.com/sonata-project/SonataAdminBundle/issues/3053) and next major [release plan](https://github.com/sonata-project/SonataAdminBundle/issues/3731), this project has been refactored regarding branching and versioning.

If you see this message, your PR concerns a patch or a minor release and is not targeting the right branch.

So I'm closing this one, but don't see it as a refusal. If you think your work is still relevant and want to continue, feel free to reopen it on the right branch (e.g. the default one).

Regards.

PR #106, opened 2015-10-20
Moved method definition to interface

core23:
This will be a bc, but I think it's better to have all public methods in the interface.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,3 +4,6 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 ### Added
+
+### Changed
+- Moved public methods from `SeoPage` to `SeoPageInterface`
--- a/Seo/SeoPage.php
+++ b/Seo/SeoPage.php
@@ -128,10 +128,7 @@ class SeoPage implements SeoPageInterface
     }
 
     /**
-     * @param string $type
-     * @param string $name
-     *
-     * @return bool
+     * {@inheritdoc}
      */
     public function hasMeta($type, $name)
     {
@@ -139,10 +136,7 @@ class SeoPage implements SeoPageInterface
     }
 
     /**
-     * @param string $type
-     * @param string $name
-     *
-     * @return $this
+     * {@inheritdoc}
      */
     public function removeMeta($type, $name)
     {
@@ -194,9 +188,7 @@ class SeoPage implements SeoPageInterface
     }
 
     /**
-     * @param string $name
-     *
-     * @return $this
+     * {@inheritdoc}
      */
     public function removeHtmlAttributes($name)
     {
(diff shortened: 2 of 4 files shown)


OskarStark:
Only if they are really needed, IMO the interface should just define the minimal amount of methods you need.

core23:
It will be more common to ``remove`` something then setting something completly to null. 

When they shouldn't be accessed outside, the methods could be private or protected.

OskarStark:
the could, but don't have to, to implement the interface

what do you say @Soullivaneuh

Soullivaneuh:
I don't know enough this bundle to say if these methods should be on interface or not.

In any way yes, this is BC break on would be merged only for major.

Any chance to have BC handling @core23?

core23:
No idea how to handle this break. You could introduce a new ``AdvancedSeoPageInterface`` but this would get messed by time...

I recently read the [symfony docs](http://symfony.com/doc/current/contributing/code/bc.html) for bc breaks. They handle this changes with care, but you were allowed to add methods to interfaces when adding a hint to the **UPGRADE** file. 

What I got from your approach would mean, that changes on interfaces are only allowed any few years when a new mayor version gets released.

Soullivaneuh:
@core23 Please rebase.

OskarStark:
@Soullivaneuh i think we can merge this in 3.x, right?

Soullivaneuh:
> @Soullivaneuh i think we can merge this in 3.x, right?

Yes, `master` **is** `3.x`.

OskarStark:
Sorry i was confused across all the bundles :neckbeard::grimacing:

Soullivaneuh:
@core23 Could you please rebase and fill the needs labels?

core23:
Done @Soullivaneuh

CHANGELOG.md
```diff
@@ -4,3 +4,6 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 ### Added
+
+### Changed
+- moved public methods in `SeoPage` to `SeoPageInterface`
```
Soullivaneuh: **M**oved

Soullivaneuh:
@core23 Your upgrade not must be on `UPGRADE-4.0.md`, not 3.0.

core23:
fixed

Soullivaneuh:
RTM. :+1: 

Feel free to merge it once CI will be green.

CHANGELOG.md
```diff
@@ -4,3 +4,6 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 ### Added
+
+### Changed
+- Moved public methods in `SeoPage` to `SeoPageInterface`
```
Soullivaneuh: Just one thing again:

`Moved public methods in`

Should not be **from** instead of in?

Issue #113, opened 2016-02-23
Breadcrumb block attributes


rYsvelain:
Hello,

I was doing this part of the documentation : https://sonata-project.org/bundles/seo/master/doc/reference/breadcrumb.html

But the BaseBreadcrumbMenuBlockService class, has the factory and the context attributes which are private. The documentation ask to extend this class, but If I want to use the factory in order to create a separate menu I can't.

It would be better to have these attributes in protected ?

Th3Mouk:
Hello ! Why U don't use the factory and context getter meantime ?

Can you provide a PR with the correction ? Thx

PR #114, opened 2016-02-23
[MOD] Attributes visibility

Th3Mouk:
https://github.com/sonata-project/SonataSeoBundle/issues/113


core23:
:-1: There are getters for this

Th3Mouk:
that was i said but why private and not protected ?

core23:
No idea. @rande created the file long time ago...

greg0ire:
Maybe because he read [this](http://fabien.potencier.org/pragmatism-over-theory-protected-vs-private.html) ?

core23:
When changing the fields to protected, we should remove the same methods for the next major release.

Th3Mouk:
This makes sense my first approach was good. Thx !

PR #118, opened 2016-03-11
Enable configuration file for each bundle

cmizzi:


greg0ire:
I'm a big advocate of small commits, but yours are related together, you should squash them.

PR #123, opened 2016-05-07
Allow SonataBlockBundle 3.0

### LINKED ISSUE #122: Add support for BlockBundle 3.0
Sonata admin 3.0 requires BlockBundle 3.0. This means the seo and admin bundle can't be installed together on Symfony 3 atm.

WouterJ:
Fixes #122 

I can't see any BC breaks between BlockBundle 2.3.11 and master, so I think 3.0-dev is already supported.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,18 +17,18 @@
         }
     ],
     "require": {
-        "symfony/framework-bundle": "~2.3|~3.0",
-        "symfony/options-resolver": "~2.3|~3.0",
+        "symfony/framework-bundle": "~2.3 || ~3.0",
+        "symfony/options-resolver": "~2.3 || ~3.0",
         "sonata-project/exporter": "~1.2,>=1.2.2",
-        "sonata-project/block-bundle": "^2.3.9",
+        "sonata-project/block-bundle": "^2.3.9 || ~3.0",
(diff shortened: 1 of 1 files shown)


Soullivaneuh:
The idea is to push major on most bundles to start with a new one but seo-bundle 2.0 looks pretty recent so we could make an exception.

Also, block bundle 3.0 is gonna be released soon. I'll ping you for a rebase after that.

Soullivaneuh:
Block bundle 3.0 is out and Travis is not started yet. This should be OK so.

composer.json
```diff
@@ -17,18 +17,18 @@
         }
     ],
     "require": {
-        "symfony/framework-bundle": "~2.3|~3.0",
-        "symfony/options-resolver": "~2.3|~3.0",
+        "symfony/framework-bundle": "~2.3 || ~3.0",
```
core23: We should change this to ``"symfony/framework-bundle": "^2.3 || ^3.0"``

WouterJ: Let's do that in another PR?

Soullivaneuh: Well, since you changed the constraint notation, I think we can have this modification too on this PR.

Soullivaneuh:
Closed for #124.

PR #161, opened 2016-06-26
Suggest HTTPlug instead of Guzzle implementation

core23:
I am targetting this branch, because this adds a new implementation without removing the existing one.

## Changelog

``` markdown
### Added
- Added support for `HTTPlug` in `TwitterEmbedTweetBlockService`
```
## Subject

Added support for an abstract http client without removing the existing old guzzle implementation.

### DIFF
--- a/UPGRADE-2.x.md
+++ b/UPGRADE-2.x.md
@@ -1,6 +1,13 @@
 UPGRADE 2.x
 ===========
 
+## TwitterEmbedTweetBlockService uses HTTPlug
+
+The `Guzzle` dependency was removed and replaced with the abstract `HTTPlug` client, so you can choose your preferred 
+http client implementation. If you have extended the `TwitterEmbedTweetBlockService` class, you need to adjust the 
+constructor signature.
+
+
 UPGRADE FROM 2.2 to 2.3
 =======================
 
--- a/composer.json
+++ b/composer.json
@@ -35,7 +35,10 @@
         "sonata-project/block-bundle": "<3.11"
     },
     "require-dev": {
-        "guzzle/guzzle": "^3.8",
+        "guzzlehttp/guzzle": "^6.0",
+        "php-http/httplug": "^1.1 || ^2.0",
+        "php-http/httplug-bundle": "^1.14",
+        "php-http/mock-client": "^0.3",
         "sonata-project/admin-bundle": "^3.31",
         "sonata-project/block-bundle": "^3.11",
         "sonata-project/core-bundle": "^3.9",
@@ -46,8 +49,9 @@
         "symfony/yaml": "^2.8 || ^3.2 || ^4.0"
     },
     "suggest": {
-        "guzzle/guzzle": "3.*",
-        "knplabs/knp-menu-bundle": "Used by the BreadcrumbMenuBuilder"
+        "knplabs/knp-menu-bundle": "Used by the BreadcrumbMenuBuilder",
+        "php-http/buzz-adapter": "Buzz HTTP client implementation for TwitterEmbedTweetBlockService",
+        "php-http/guzzle6-adapter": "Guzzle HTTP client implementation for TwitterEmbedTweetBlockService"
     },
     "config": {
         "sort-packages": true
--- a/composer.json
+++ b/composer.json
@@ -38,7 +38,7 @@
         "guzzlehttp/guzzle": "^6.0",
         "php-http/httplug": "^1.1 || ^2.0",
         "php-http/httplug-bundle": "^1.13",
-        "php-http/mock-client": "^0.3",
+        "php-http/mock-client": "^1.1",
         "sonata-project/admin-bundle": "^3.31",
         "sonata-project/block-bundle": "^3.11",
         "sonata-project/core-bundle": "^3.9",
--- a/composer.json
+++ b/composer.json
@@ -37,7 +37,7 @@
     "require-dev": {
         "guzzlehttp/guzzle": "^6.0",
         "php-http/httplug": "^1.1 || ^2.0",
-        "php-http/httplug-bundle": "^1.14",
+        "php-http/httplug-bundle": "^1.13",
         "php-http/mock-client": "^0.3",
         "sonata-project/admin-bundle": "^3.31",
         "sonata-project/block-bundle": "^3.11",
--- a/docs/reference/social_blocks.rst
+++ b/docs/reference/social_blocks.rst
@@ -49,5 +49,12 @@ The ``sonata.seo.block.twitter.embed`` allows you to embed a tweet by giving its
 or embed content. Please refer to `Embedded Tweets doc <https://dev.twitter.com/docs/embedded-tweets>`_
 and `OEmbed API doc <https://dev.twitter.com/docs/api/1/get/statuses/oembed>`_ for a full understanding.
 
-The block service allows you to ask the API if you've given an ID or a URL (you'll need to install the
-Guzzle library for this: ``composer require guzzle/guzzle``).
+The block service allows you to ask the API if you've given an ID or a URL .
+
+You'll need to install an HTTPlug client to request the Twitter API, for example the Guzzle adapter:
+
+.. code-block:: bash
+
+    composer require php-http/guzzle6-adapter
+
+More information can be found [here](https://php-http.readthedocs.io/en/latest/httplug/users.html).
--- a/src/Block/Social/TwitterEmbedTweetBlockService.php
+++ b/src/Block/Social/TwitterEmbedTweetBlockService.php
@@ -54,13 +54,7 @@ class TwitterEmbedTweetBlockService extends BaseTwitterButtonBlockService
      */
     private $messageFactory;
 
-    /**
-     * @param string          $name
-     * @param EngineInterface $templating
-     * @param HttpClient      $httpClient
-     * @param MessageFactory  $messageFactory
-     */
-    public function __construct($name, EngineInterface $templating, HttpClient $httpClient = null, MessageFactory $messageFactory = null)
+    public function __construct(?string $name, EngineInterface $templating, HttpClient $httpClient = null, MessageFactory $messageFactory = null)
     {
         parent::__construct($name, $templating);
 
@@ -68,10 +62,7 @@ class TwitterEmbedTweetBlockService extends BaseTwitterButtonBlockService
         $this->messageFactory = $messageFactory;
     }
 
-    /**
-     * {@inheritdoc}
-     */
-    public function execute(BlockContextInterface $blockContext, Response $response = null)
+    public function execute(BlockContextInterface $blockContext, Response $response = null): Response
     {
         return $this->renderResponse($blockContext->getTemplate(), [
             'block' => $blockContext->getBlock(),
--- a/src/Block/Social/TwitterEmbedTweetBlockService.php
+++ b/src/Block/Social/TwitterEmbedTweetBlockService.php
@@ -54,8 +54,12 @@ class TwitterEmbedTweetBlockService extends BaseTwitterButtonBlockService
      */
     private $messageFactory;
 
-    public function __construct(?string $name, EngineInterface $templating, HttpClient $httpClient = null, MessageFactory $messageFactory = null)
-    {
+    public function __construct(
+        ?string $name,
+        EngineInterface $templating,
+        HttpClient $httpClient = null,
+        MessageFactory $messageFactory = null
+    ) {
         parent::__construct($name, $templating);
 
         $this->httpClient = $httpClient;
--- a/src/Block/Social/TwitterEmbedTweetBlockService.php
+++ b/src/Block/Social/TwitterEmbedTweetBlockService.php
@@ -13,12 +13,16 @@ declare(strict_types=1);
 
 namespace Sonata\SeoBundle\Block\Social;
 
-use Guzzle\Http\Exception\CurlException;
+use GuzzleHttp\Exception\GuzzleException;
+use Http\Client\Exception;
+use Http\Client\HttpClient;
+use Http\Message\MessageFactory;
 use Sonata\AdminBundle\Form\FormMapper;
 use Sonata\BlockBundle\Block\BlockContextInterface;
 use Sonata\BlockBundle\Model\BlockInterface;
 use Sonata\CoreBundle\Form\Type\ImmutableArrayType;
 use Sonata\CoreBundle\Model\Metadata;
+use Symfony\Bundle\FrameworkBundle\Templating\EngineInterface;
 use Symfony\Component\Form\Extension\Core\Type\CheckboxType;
 use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
 use Symfony\Component\Form\Extension\Core\Type\IntegerType;
@@ -41,36 +45,37 @@ class TwitterEmbedTweetBlockService extends BaseTwitterButtonBlockService
     public const TWEET_ID_PATTERN = '%^([0-9]*)$%';
 
     /**
+     * @var HttpClient|null
+     */
+    private $httpClient;
+
+    /**
+     * @var MessageFactory|null
+     */
+    private $messageFactory;
+
+    /**
+     * @param string          $name
+     * @param EngineInterface $templating
+     * @param HttpClient      $httpClient
+     * @param MessageFactory  $messageFactory
+     */
+    public function __construct($name, EngineInterface $templating, HttpClient $httpClient = null, MessageFactory $messageFactory = null)
+    {
+        parent::__construct($name, $templating);
+
+        $this->httpClient = $httpClient;
+        $this->messageFactory = $messageFactory;
+    }
+
+    /**
      * {@inheritdoc}
      */
     public function execute(BlockContextInterface $blockContext, Response $response = null)
     {
-        $tweet = $blockContext->getSetting('tweet');
-
-        if (($uriMatched = preg_match(self::TWEET_URL_PATTERN, $tweet))
-            || preg_match(self::TWEET_ID_PATTERN, $tweet)) {
-            // We matched an URL or an ID, we'll need to ask the API
-            if (false === class_exists('Guzzle\Http\Client')) {
-                throw new \RuntimeException('The guzzle http client library is required to call the Twitter API. Make sure to add guzzle/guzzle to your composer.json.');
-            }
-
-            // TODO cache API result
-            $client = new \Guzzle\Http\Client();
-            $client->setConfig(['curl.options' => [CURLOPT_CONNECTTIMEOUT_MS => 1000]]);
-
-            try {
-                $request = $client->get($this->buildUri($uriMatched, $blockContext->getSettings()));
-                $apiTweet = json_decode($request->send()->getBody(true), true);
-
-                $tweet = $apiTweet['html'];
-            } catch (CurlException $e) {
-                // log error
-            }
-        }
-
         return $this->renderResponse($blockContext->getTemplate(), [
             'block' => $blockContext->getBlock(),
-            'tweet' => $tweet,
+            'tweet' => $this->loadTweet($blockContext),
         ], $response);
     }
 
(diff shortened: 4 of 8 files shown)


Block/Social/TwitterEmbedTweetBlockService.php
```diff
@@ -204,4 +208,49 @@ protected function buildUri($uriMatched, array $settings)
...
     }
+
+    /**
+     * Loads twitter tweet.
+     *
+     * @param array $settings
+     *
+     * @return string
+     */
+    final protected function loadTweet($settings)
```
OskarStark: could we typehint `$settings` as an array?

Block/Social/TwitterEmbedTweetBlockService.php
```diff
@@ -204,4 +208,49 @@ protected function buildUri($uriMatched, array $settings)
...
+                $client = new \Guzzle\Http\Client();
+            $client->setConfig(array('curl.options' => array(CURLOPT_CONNECTTIMEOUT_MS => 1000)));
+
+            try {
+                $request = $client->get($this->buildUri($uriMatched, $settings));
+                $apiTweet = json_decode($request->send()->getBody(true), true);
+
+                return $apiTweet['html'];
+            } catch (CurlException $e) {
+                // log error
```
OskarStark: ?

DependencyInjection/Configuration.php
```diff
@@ -62,6 +62,15 @@ public function getConfigTreeBuilder()
                         ->end()
                     ->end()
                 ->end()
+                ->children()
+                    ->arrayNode('http')
+                        ->addDefaultsIfNotSet()
+                        ->children()
+                            ->scalarNode('client')->defaultValue('httplug.client.default')->end()
```
OskarStark: could you please add `->info()` nodes to both config options?

Block/Social/TwitterEmbedTweetBlockService.php
```diff
@@ -204,4 +208,49 @@ protected function buildUri($uriMatched, array $settings)
...
+                // NEXT_MAJOR: Remove the old guzzle implementation
+
+                // We matched an URL or an ID, we'll need to ask the API
+                if (class_exists('Guzzle\Http\Client') === false) {
+                    throw new \RuntimeException(
+                        'The guzzle http client library is required to call the Twitter API. Make sure to add guzzle/guzzle to your composer.json.'
+                    );
+                }
+
+                // TODO cache API result
```
OskarStark: ?

Soullivaneuh:
The diff is quite hard to read for `TwitterEmbedTweetBlockService.php`.

Could you please split commits to make it easier to read?

cc @greg0ire

Block/Social/TwitterEmbedTweetBlockService.php
core23: This is just the original Guzzle code without any change. This will be removed with the next major release.

Block/Social/TwitterEmbedTweetBlockService.php
core23: Original Guzzle code, without any modification

core23:
Splitted

core23:
Rebased

core23:
Can you have a look at this @sagikazarmark

Block/Social/TwitterEmbedTweetBlockService.php
```diff
@@ -35,36 +38,37 @@ class TwitterEmbedTweetBlockService extends BaseTwitterButtonBlockService
...
-            try {
-                $request = $client->get($this->buildUri($uriMatched, $blockContext->getSettings()));
-                $apiTweet = json_decode($request->send()->getBody(true), true);
+    /**
+     * @param string          $name
+     * @param EngineInterface $templating
+     * @param HttpClient      $httpClient
+     * @param MessageFactory  $messageFactory
+     */
+    public function __construct($name, EngineInterface $templating, HttpClient $httpClient = null, MessageFactory $messageFactory = null)
```
sagikazarmark: From my point of view, null value makes sense if you want to/can fallback to a default client. I understand that this is for BC reason, however, when you remove the old guzzle code, you can choose from two options:

1. Remove optional (BC break)
2. Leave optional and use php-http/discovery

Option two is kind of a BC break too, since it modifies the "default" value. Therefore I would probably do the following: implement php-http/discovery now, BUT catch any exceptions coming out of it. That way you don't change the behavior of the code (too much, since you will have to remove the try catch, but the dependencies will require to install an HTTP Client anyway, so you shouldn't receive any unexpected errors from the code).

The other reason why implementing php-http/discovery could be a good idea, because it would use any already existing, discoverable clients.

Of course, if you don't want any fallback logic to be used in the next major, just ignore my comment and leave it as is. Since this is a bundle anyway, it will probably be used as a service which requires configuration anyway.

Block/Social/TwitterEmbedTweetBlockService.php
```diff
@@ -204,4 +208,53 @@ protected function buildUri($uriMatched, array $settings)
...
+            if (class_exists('Guzzle\Http\Client') === false) {
+                throw new \RuntimeException(
+                    'The guzzle http client library is required to call the Twitter API. Make sure to add guzzle/guzzle to your composer.json.'
+                );
+            }
+
+            @trigger_error('The direct Guzzle implementation is deprecated since 2.x and will be removed with the next major release.', E_USER_DEPRECATED);
+
+            // TODO cache API result
+            $client = new \Guzzle\Http\Client();
```
sagikazarmark: Maybe you could write a stupid simple Guzzle 3 wrapper which implements HttpClient interface. It would require to add httplug as a dependency though, but it would make the code a bit more consistent IMO.

Resources/doc/reference/social_blocks.rst
```diff
@@ -45,4 +45,12 @@ Twitter Embed Tweet
...
\ No newline at end of file
+The block service allows you to ask the API if you've given an ID or a URL.
+
+You'll need to install a HTTPlug client to request the Twitter API, for example the Guzzle adapter:
+
+.. code-block:: bash
+
+    $ composer require php-http/guzzle6-adapter`
+
+More information can be found [here](http://httplug.io/).
```
sagikazarmark: Hm, we don't have an official recommendation about linking to the website or the docs (http://docs.php-http.org/en/latest/).

WDYT?

/cc @dbu @Nyholm

Nyholm: I usually link to this page: http://php-http.readthedocs.io/en/latest/httplug/users.html

It explains the basics of HTTPlug and how you use it.

sagikazarmark:
@core23 Added a few points, but I don't think it should keep you from merging. :+1:

core23:
Thanks for your review @sagikazarmark  ❤️ 

Will work on this the following days.

DependencyInjection/SonataSeoExtension.php
```diff
@@ -187,4 +188,14 @@ protected function fixConfiguration(array $config)
...
         return $config;
     }
+
+    /**
+     * @param ContainerBuilder $container
+     * @param array            $config
+     */
+    private function configureHttpClient(ContainerBuilder $container, array $config)
+    {
+        $container->setAlias('sonata.seo.http.client', $config['http']['client']);
```
dbu: did you test that this does not lead to an error? afaik the on-invalid only works if the service itself is not found, but here you create an alias which is found, and then get a container error about the alias not found. maybe you need a compiler pass for this: set the value as container parameter, in the pass check if the configured service actually exists and only add the alias in this case. the blocks.xml file could stay as is, on-invalid without the alias will just pass in null

Resources/doc/reference/social_blocks.rst
dbu: yep, that users page is meant as the entry point for people needing httplug for a library they use. additionally, a link to http://php-http.readthedocs.io/en/latest/integrations/symfony-bundle.html would also make sense, as this is the doc for a bundle and it assumes people to configure the client as a service.

dbu: or not - i think we should take care of that on the httplug doc side: https://github.com/php-http/documentation/pull/146

OskarStark:
stuck build @Soullivaneuh

greg0ire:
@OskarStark it's fixed

OskarStark:
@core23 please check travis, thank you!

Block/Social/TwitterEmbedTweetBlockService.php
core23: A good idea in general, but...

👎 we would introduce a new wrapper class and mark it as deprecated, because we would remove it with the next major release.

Block/Social/TwitterEmbedTweetBlockService.php
core23: We want a BC version, so the parameters can be null at the moment. This logic will be changed and the concrete Guzzle implementation will be removed.

Can you give me an example how to use the discover mechanism? Or is this only required for the next major? Sorry, it's my first try with HTTPlug.

Block/Social/TwitterEmbedTweetBlockService.php
sagikazarmark: I don't see it as a problem. You would have a more stable code base now without deprecations in it. IMO it's better than having deprecated legacy code.

core23: What do you think @greg0ire @Soullivaneuh ? Should we introduce a new wrapper class which will be removed with 4.0?

greg0ire: I don't think there is a point in triggering a deprecation if people can't do anything to use "the new way", which here would be to use httplug, IMO the deprecation error should ask people to install httplug, and if they do, this piece of code should use it. It they don't, we should use the old client.

> It would require to add httplug as a dependency though, but it would make the code a bit more consistent IMO.

I think we shouldn't require httplug, that would be a BC-break. Instead, we should detect it, detect the presence of a modern implementation of an http client, and fallback to the old code (so, no wrapper, because if httplug is not installed, the code must work). The deprecation should ask to install httplug AND an http client implementation IMO.

dbu: requiring httplug without requiring an implementation is no BC break. i would do that, write the guzzle 3 adapter and use that as fallback if the discovery finds no adapter. the benefit is that the rest of the code can be consistent about only handling the Httplug client interface and not distinguish between plain guzzle or httplug.

and i agree we should only trigger a deprecation notice if we have to fall back to guzzle 3, not in other cases.

Block/Social/TwitterEmbedTweetBlockService.php
sagikazarmark: No problem, I am happy to help.

The discovery is a convenience layer which helps detecting installed clients. OOTB the official clients (implemented by us) are supported, but it can be configured to use [Puli](http://puli.io) which allows to discover custom clients too.

See an example of discovery here: http://docs.php-http.org/en/latest/discovery.html#http-client-discovery

If the discovery is used and no resource (client, message factory, etc) is found an exception is thrown, which is fine, as your code will probably not work without it.

In your case, you would have to catch these exceptions to preserve BC for the moment and fall back to guzzle 3.

I would recommend using the discovery now (if you want to use it), because that way if any of our clients is installed, the discovery will find it and the consuming code can use it instead of guzzle 3.

But as said, this is a Bundle and DependencyInjection is always preferred over discovery. That said, in this case it could be useful because of the reasons mentioned above.

Block/Social/TwitterEmbedTweetBlockService.php
sagikazarmark: Exactly my thoughts

OskarStark:
any news here @core23 ?

core23:
This should work now, except for old PHP versions.  HTTPlug requires at least PHP 5.5 😞

greg0ire:
Argh… it means you need to `require --dev` only for some jobs, and skip some tests otherwise. Not sure how to achieve that with dev-kit…

core23:
> Argh… it means you need to `require --dev` only for some jobs, and skip some tests otherwise. Not sure how to achieve that with dev-kit…

Exactly. That's the problem. IMHO the only was to archive this is to drop PHP <=5.5 or use dirty hacks for travis.

greg0ire:
Can't think of any other solution indeed. Maybe we could stop running the Travis jobs for old versions of php? That way we're not forced to drop old versions…

core23:
Be could add this to `Allowed Failures`, but who would check if our software is still running on old PHP versions? Or should we just ignore that fact?

greg0ire:
I really don't know… maybe the dirty Travis hacks are way to go. It would imply adding a hook system to dev-kit (check for files in `.travis` and execute them). Maybe try doing it without dev-kit to see if it's easy?

dbu:
anything before 5.6 does not even get security fixes and should not be used
anymore. imho we should do a new minor version that is restricted to 5.6 /
7. people stuck on legacy can keep the current minor version. we should not
encourage bad practice by supporting dangerous php versions

core23:
My words @dbu 

But long time ago we decided dropping PHP versions is a BC break, so we are stuck at the moment.

greg0ire:
> But long time ago we decided dropping PHP versions is a BC break

Maybe it's time to change that? Can you open a dev-kit issue about that?

core23:
We can't merge this, because we have old PHP support :(

So should we close this?

dbu:
or make this a pull request on the next major version?

composer.json
```diff
@@ -36,6 +36,8 @@
     },
     "require-dev": {
         "guzzle/guzzle": "3.*",
```
covex-nn: I suggest to replace this package with `guzzlehttp/guzzle`, since `guzzle/guzzle` was abbadoned and no longer maintained, see https://packagist.org/packages/guzzle/guzzle#v3.8.1

UPGRADE-2.x.md
```diff
@@ -1,6 +1,12 @@
 UPGRADE 2.x
 ===========
 
+## TwitterEmbedTweetBlockService uses HTTPlug
+
+The `Guzzle` dependency was removed and replaced with the abstract `HTTPlug` client, so you can choose your prefered 
+http client implementation. If you need have extended this class, you need to adopt the new constructor signature.
```
dbu: If you have extended the class `TwitterEmbedTweetBlockService`, you need to adjust the constructor signature.

src/Block/Social/TwitterEmbedTweetBlockService.php
```diff
@@ -208,4 +212,56 @@ protected function buildUri($uriMatched, array $settings)
...
+
+            return $apiTweet['html'];
+        }
+
+        // NEXT_MAJOR: Remove the old guzzle implementation
+
+        // We matched an URL or an ID, we'll need to ask the API
+        if (false === class_exists('GuzzleHttp\Client')) {
+            throw new \RuntimeException(
+                'The guzzle http client library is required to call the Twitter API. Make sure to add guzzle/guzzle to your composer.json.'
```
dbu: HTTPlug or Guzzle is required to... Add php-http/httplug-bundle or guzzlehttp/guzzle to...

src/Block/Social/TwitterEmbedTweetBlockService.php
```diff
@@ -208,4 +212,56 @@ protected function buildUri($uriMatched, array $settings)
...
+    private function loadTweet(BlockContextInterface $blockContext)
+    {
+        $uriMatched = preg_match(self::TWEET_URL_PATTERN, $blockContext->getSetting('tweet'));
+
+        if (!$uriMatched || !preg_match(self::TWEET_ID_PATTERN, $blockContext->getSetting('tweet'))) {
+            return null;
+        }
+
+        if (null !== $this->httpClient && null !== $this->messageFactory) {
+            $response = $this->httpClient->sendRequest(
```
dbu: should there be a try-catch as well, to handle web request exceptions?

UPGRADE-2.x.md
```diff
@@ -1,6 +1,12 @@
 UPGRADE 2.x
 ===========
 
+## TwitterEmbedTweetBlockService uses HTTPlug
+
+The `Guzzle` dependency was removed and replaced with the abstract `HTTPlug` client, so you can choose your prefered
```
greg0ire: "preferred"

UPGRADE-2.x.md
```diff
@@ -1,6 +1,12 @@
 UPGRADE 2.x
 ===========
 
+## TwitterEmbedTweetBlockService uses HTTPlug
+
+The `Guzzle` dependency was removed and replaced with the abstract `HTTPlug` client, so you can choose your prefered 
+http client implementation. If you need have extended this class, you need to adjust the constructor signature.
```
greg0ire: "If you have"

greg0ire: Please say which class explicitely

docs/reference/social_blocks.rst
```diff
@@ -45,4 +45,12 @@ Twitter Embed Tweet
 
 The ``sonata.seo.block.twitter.embed`` allows you to embed a tweet by giving its id, URL or embed content. Please refer to `Embedded Tweets doc <https://dev.twitter.com/docs/embedded-tweets>`_ and `OEmbed API doc <https://dev.twitter.com/docs/api/1/get/statuses/oembed>`_ for a full understanding.
 
-The block service allows you to ask the API if you've given an ID or a URL (you'll need to install the Guzzle library for this: ``composer require guzzle/guzzle``).
+The block service allows you to ask the API if you've given an ID or a URL .
+
+You'll need to install a HTTPlug client to request the Twitter API, for example the Guzzle adapter:
```
greg0ire: "an"

composer.json
```diff
@@ -46,8 +48,9 @@
         "symfony/yaml": "^2.8 || ^3.2 || ^4.0"
     },
     "suggest": {
-        "guzzle/guzzle": "3.*",
-        "knplabs/knp-menu-bundle": "Used by the BreadcrumbMenuBuilder"
+        "knplabs/knp-menu-bundle": "Used by the BreadcrumbMenuBuilder",
+        "php-http/buzz-adapter": "Buzz HTTP client implementation for TwitterEmbedTweetBlockService",
+        "php-http/guzzle6-adapter": "Guzzle HTTP client implementation for TwitterEmbedTweetBlockService"
```
greg0ire: Shouldn't this be in `require-dev` for BC?

OskarStark:
Is this ready to merge @core23 ?

dbu:
"PSR-18 HTTP Client" has been released a few days ago. it is the official standard that evolved out of HTTPlug. i think it would make sense to change this PR to use PSR-18 instead. the php-http group is working on updating the guzzle adapter to support PSR-18 in addition to httplug (and i guess guzzle and buzz will soon implement PSR-18 directly)

OskarStark:
Sure we should evolve, but if this PR is ready and working, we should merge it and work from this point of code to implement PSR-18 or do I miss sth.?

dbu:
the thing is, we add dependencies here that we then would drop soon again to change things some more. people setting this bundle up with a httplug client would need to change again to provide a psr-18 client implementation, so its another BC break. as PSR-18 is now out, i think we could jump that step. the code should not require any changes actually, as the signature is the same (safe for the return type declaration in PSR-18)

dbu:
hm, okay. if you want to keep supporting PHP 5.6 you need to use HTTPlug for now. PSR-18 does not support PHP 5.6 (because that PHP version is obsolete). if you want to merge as is, go ahead, i am not saying its a terrible idea, just that there are some bumps to be aware of.

OskarStark:
@core23 please give us some feedback

src/DependencyInjection/SonataSeoExtension.php
```diff
@@ -169,4 +170,14 @@ protected function fixConfiguration(array $config)
 
         return $config;
     }
+
+    /**
+     * @param ContainerBuilder $container
+     * @param array            $config
```
OskarStark: could be removed

src/Block/Social/TwitterEmbedTweetBlockService.php
```diff
@@ -41,36 +45,37 @@ class TwitterEmbedTweetBlockService extends BaseTwitterButtonBlockService
...
 
-            try {
-                $request = $client->get($this->buildUri($uriMatched, $blockContext->getSettings()));
-                $apiTweet = json_decode($request->send()->getBody(true), true);
+    /**
+     * @param string          $name
+     * @param EngineInterface $templating
+     * @param HttpClient      $httpClient
+     * @param MessageFactory  $messageFactory
+     */
```
greg0ire: Please remove the php doc and add a parameter type hint

src/Block/Social/TwitterEmbedTweetBlockService.php
```diff
@@ -41,36 +45,37 @@ class TwitterEmbedTweetBlockService extends BaseTwitterButtonBlockService
...
-            }
-        }
+        $this->httpClient = $httpClient;
+        $this->messageFactory = $messageFactory;
+    }
 
+    /**
+     * {@inheritdoc}
+     */
+    public function execute(BlockContextInterface $blockContext, Response $response = null)
```
greg0ire: Please add a return type hint

src/Block/Social/TwitterEmbedTweetBlockService.php
```diff
@@ -41,36 +45,28 @@ class TwitterEmbedTweetBlockService extends BaseTwitterButtonBlockService
...
-            $client->setConfig(['curl.options' => [CURLOPT_CONNECTTIMEOUT_MS => 1000]]);
+    /**
+     * @var MessageFactory|null
+     */
+    private $messageFactory;
 
-            try {
-                $request = $client->get($this->buildUri($uriMatched, $blockContext->getSettings()));
-                $apiTweet = json_decode($request->send()->getBody(true), true);
+    public function __construct(?string $name, EngineInterface $templating, HttpClient $httpClient = null, MessageFactory $messageFactory = null)
```
greg0ire:
```suggestion
    public function __construct(
        ?string $name,
        EngineInterface $templating,
        ?HttpClient $httpClient = null,
        ?MessageFactory $messageFactory = null
    )
```

PR #163, opened 2016-06-29
Remove internal test classes from autoloader

core23:
I am targetting this branch, because some bundles use the test classes.

Related to https://github.com/sonata-project/dev-kit/issues/179

## Changelog

```markdown
### Removed
- internal test classes are now excluded from the autoloader
```

## Subject

Tests for internal components shouldn't be loaded in production or extended by other bundles.

### DIFF
--- a/UPGRADE-3.0.md
+++ b/UPGRADE-3.0.md
@@ -10,3 +10,9 @@ If you have implemented a custom seo page, you must adapt the signature of the f
  * `hasHeadAttribute`
  * `removeLangAlternate`
  * `hasLangAlternate`
+
+### Tests
+
+All files under the ``Tests`` directory are now correctly handled as internal test classes. 
+You can't extend them anymore, because they are only loaded when running internal tests. 
+More information can be found in the [composer docs](https://getcomposer.org/doc/04-schema.md#autoload-dev).
--- a/composer.json
+++ b/composer.json
@@ -37,8 +37,14 @@
         "knplabs/knp-menu-bundle": "Used by the BreadcrumbMenuBuilder"
     },
     "autoload": {
-        "psr-4": { "Sonata\\SeoBundle\\": "" }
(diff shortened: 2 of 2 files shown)


OskarStark:
LGTM 👍

greg0ire:
Thanks @core23 !

Soullivaneuh:
Minor, not major! Already said in https://github.com/sonata-project/dev-kit/issues/179#issuecomment-229347730

greg0ire:
> I am targetting this branch, because some bundles use the test classes.

So, it should be major, right? Because if we do that, related bundles won't be able to use the tests anymore (without doing calls to `require_once`)

Soullivaneuh:
> So, it should be major, right? Because if we do that, related bundles won't be able to use the tests anymore 

So bump the internal constraint and/or fix the tests.

greg0ire:
> So bump the internal constraint and/or fix the tests.

Which is touching your code. Which mean there was a BC break. Which means this should be major. No?

Soullivaneuh:
> Which is touching your code. Which mean there was a BC break.

No.

http://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api

Also: https://github.com/sonata-project/dev-kit/issues/179#issuecomment-229599354

greg0ire:
> http://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api

This has nothing to do with the issue at hand IMO.

Here we have SonataSeoBundle. If we change something in this bundle and it means people (and in people I include "other bundles") have to change their code (be it fix their tests or whatever) if they upgrade, that means the change is a BC-break. And every reasonable BC-break should trigger a major release.

So in this case, unless we consider it's not a reasonable BC-break (meaning we consider people should not rely on the test classes being public), it should be major.

PR #181, opened 2017-01-16
removed useless class comment

OskarStark:
I am targetting this branch, because this is BC.

## Subject
Only a PHP Doc fix


OskarStark:
fixed the CS

PR #187, opened 2017-01-24
Fix wrong default value form option for sonata.seo.block.twitter.embed block

gremo:
I am targetting this branch, because this is a bugfix.

## Changelog

```markdown
### Changed
- Default value for `lang` to `null `instead of array of languages
```


core23:
Do we need the `lang` option anyway? A `null` value seems wrong too.

greg0ire:
Why? What is wrong?

greg0ire:
Found the explanation in the issue, but the explanation should be in the commit message.

PR #204, opened 2017-05-08
Fix sitemap.rst documentation

Vesax:
I am targeting this branch, because it's documentation fix.

## Changelog

```markdown
### Fixed
- Invalid sitemap query example
```
## Subject
Without checking for `publication_date_end` we get all previous publications for page, even if they are no longer active


greg0ire:
Thanks @Vesax!

PR #205, opened 2017-05-13
Fix for Twig 2

Filoz:
I am targeting this branch, because branch 2.x supports Twig 2.

## Changelog

```markdown

### Changed
 - `SonataSeoBundle/Resources/views/Block/breadcrumb.html.twig` changed "sameas" in "same as"

```

## Subject

In twig 2 "sameas" bacames "same as" ( https://twig.sensiolabs.org/doc/2.x/tests/sameas.html)

"same as" works also in in Twig 1 because it is an alias of "sameas" (https://twig.sensiolabs.org/doc/1.x/tests/sameas.html)

### DIFF
--- a/Resources/views/Block/breadcrumb.html.twig
+++ b/Resources/views/Block/breadcrumb.html.twig
@@ -15,7 +15,7 @@ file that was distributed with this source code.
     {% set translation_domain = item.getExtra('translation_domain', 'SonataSeoBundle') %}
     {% if options.allow_safe_labels and item.extra('safe_label', false) %}
         {{- item.label|raw -}}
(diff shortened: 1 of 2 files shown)


Filoz:
I found a little "problem":

same as became an alias of "sameas" in Twig 1.14.2 ; composer of SeoBundle 2.x has as min version of Twig 1.12.

Can I update composer with version 1.14.2?

greg0ire:
Yes please so that it's ok

Filoz:
done :)

greg0ire:
Thanks @Filoz!

PR #220, opened 2017-10-02
Move block bundle back

ElectricMaxxx:
I am targeting this branch, because  the optional dependency was added here.

<!-
## Changelog

```markdown

### Changed

- moved `sonata-project/block-bundle` back to require-dev
```
# Subject

When block-bundle is required it forces me also to set a configuration for `sonata_block.default_context`. As the block usage is optional we should move this requiredment back to require-dev


ElectricMaxxx:
List of commits say: i need to rebase on 2.x

composer.json
```diff
@@ -44,7 +44,7 @@
      },
     "extra": {
         "branch-alias": {
-            "dev-master": "2.x-dev"
+            "dev-master": "3.x-dev"
```
core23: This is wrong

PR #221, opened 2017-10-02
move block-bundle back to require-dev

ElectricMaxxx:
I am targeting this branch, because  the optional dependency was added here.

## Changelog

```markdown

### Changed

- moved `sonata-project/block-bundle` back to require-dev
```
# Subject

When block-bundle is required it forces me also to set a configuration for `sonata_block.default_context`. As the block usage is optional we should move this requiredment back to require-dev

### DIFF
--- a/UPGRADE-2.x.md
+++ b/UPGRADE-2.x.md
@@ -1,6 +1,12 @@
 UPGRADE 2.x
 ===========
 
+UPGRADE FROM 2.2.0 to 2.2.1
+===========================
+
+`sonata-project/block-bundle` is not required anymore. If you want to use seo friendly blocks,
+you should require it on your own by calling `composer require sonata-project/block-bundle`.
+
 UPGRADE FROM 2.0 to 2.1
 =======================
 
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,6 @@
         "symfony/framework-bundle": "^2.3 || ^3.0",
         "symfony/options-resolver": "^2.3 || ^3.0",
         "sonata-project/exporter": "^1.2.2",
-        "sonata-project/block-bundle": "^3.2",
         "twig/twig": "^1.14.2 || ^2.0"
     },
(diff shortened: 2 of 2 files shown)


ElectricMaxxx:
@core23 let's take this one :-) 
please release that after merge.

greg0ire:
You could just have force pushed, you know ;)

core23:
Is this a bugfix? Otherwise we need an upgrade note

ElectricMaxxx:
I know, but something was lost on a slow WiFi from train to work (i live in germany, digital transformation still not there)

ElectricMaxxx:
bug-fix, depends on the definition. what would you say?

core23:
If the deps was added by mistake some versions ago, I would handle this a bugfix.

I would add a short note at the upgrade file to reduce future issues.

> I know, but something was lost on a slow WiFi from train to work (i live in germany, digital transformation still not there)

I know, i saw some of your talks last year at the code.talks ^^

greg0ire:
Since when do bugfixes require upgrade notes?

greg0ire:
Or did you mean BC break?

core23:
> Or did you mean BC break?

Maybe a soft BC break?

ElectricMaxxx:
just give me a hint, what to add and where, when you are ready :-)

greg0ire:
It's kinda BC-breaky, so maybe an upgrade note would be in order. Please add sth to this file: https://github.com/sonata-project/SonataSeoBundle/blob/2.x/UPGRADE-2.x.md

greg0ire:
Thanks @ElectricMaxxx !

Soullivaneuh:
:warning: 

- No release label here. Again, it was going to be ignored on the release.
- And upgrade note should not concern patches. The vhttps://github.com/sonata-project/SonataSeoBundle/pull/221/filesersion should not be specified. This is not a patch but a minor.

PR #228, opened 2017-10-15
Drop old versions of sf

jordisala1991:

There is a separate commit for sort composer packages, because it is applied to all bundles, except this (for the moment is the only I saw without it)

## Changelog

```markdown
### Removed
- Support for `^3.0` and `^3.1` versions of `framework-bundle` and `options-resolver`.
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,19 +18,19 @@
     ],
     "require": {
         "php": "^7.1",
+        "sonata-project/exporter": "^1.2.2",
         "symfony/framework-bundle": "^2.8 || ^3.0",
         "symfony/options-resolver": "^2.8 || ^3.0",
-        "sonata-project/exporter": "^1.2.2",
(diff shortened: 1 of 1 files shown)


Soullivaneuh:
@jordisala1991 Removing version support is not pedantic at all. Please be careful and don't hesitate to ask if you are not sure.

Soullivaneuh:
Seems to be already done on #226

jordisala1991:
it was pedantic on purpose, because it was modifying #226. Now it is minor, but this should have been merged before release.

Soullivaneuh:
> Now it is minor, but this should have been merged before release.

You can't determine a label based on the probable release. A release can be done at any time. :wink:

jordisala1991:
@Soullivaneuh accurate enough?

PR #248, opened 2017-12-04
Removed default title

core23:
I am targeting this branch, because this is a bugfix.

## Changelog

```markdown
### Changed
 - Changed default title to `Project name`
```

## Subject

There's no reason for `Sonata Project` as the default SEO title.

### DIFF
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -48,7 +48,8 @@ class Configuration implements ConfigurationInterface
                             ->end()
                         ->end()
                         ->scalarNode('separator')->defaultValue(' - ')->end()
-                        ->scalarNode('title')->defaultValue('Sonata Project')->end()
+                        // NEXT_MAJOR: Make this field required
+                        ->scalarNode('title')->defaultValue('Project name')->end()
                     ->end()
                 ->end()
                 ->arrayNode('sitemap')
--- a/tests/DependencyInjection/ConfigurationTest.php
+++ b/tests/DependencyInjection/ConfigurationTest.php
@@ -81,7 +81,7 @@ class ConfigurationTest extends TestCase
                 'head' => [],
                 'metas' => [],
                 'separator' => ' - ',
-                'title' => 'Sonata Project',
+                'title' => 'Project name',
             ],
             'sitemap' => [
                 'doctrine_orm' => [],


src/DependencyInjection/Configuration.php
```diff
@@ -48,7 +48,7 @@ public function getConfigTreeBuilder()
                             ->end()
                         ->end()
                         ->scalarNode('separator')->defaultValue(' - ')->end()
-                        ->scalarNode('title')->defaultValue('Sonata Project')->end()
+                        ->scalarNode('title')->defaultValue('')->end()
```
OskarStark: but then we should remove the `->defaultValue('')`, too

greg0ire: How about changing it to "Change me, moron!"

Soullivaneuh:
Not a big fan. What is the issue with the default value? Having this appearing tells the user to change it.

core23:
If we have a default title, it should have a *useful* value. **Sonata project** isn't useful for the most websites.

src/DependencyInjection/Configuration.php
```diff
@@ -48,7 +48,7 @@ public function getConfigTreeBuilder()
                             ->end()
                         ->end()
                         ->scalarNode('separator')->defaultValue(' - ')->end()
-                        ->scalarNode('title')->defaultValue('Sonata Project')->end()
+                        ->scalarNode('title')->end()
```
greg0ire: `cannotBeEmpty`? Also, shouldn't some markup be removed from the page?
If we are going this route and not putting some Lorem Ipsum everywhere, I think the logo should be removed too, right?

OskarStark:
> If we have a default title, it should have a useful value. Sonata project isn't useful for the most websites.

Switching it to "Project Name"

greg0ire:
People would indeed be more likely to change it if you name it that way

core23:
What about `Homepage` as the default page title?

greg0ire:
@core23 this shows up on every page, right? I really prefer @OskarStark 's proposal. That's something people should customize if they have many backends that use sonata. They could even customize it for other envs like staging

jordisala1991:
I also prefer @OskarStark proposal

greg0ire:
Thanks @core23 !

PR #269, opened 2018-03-21
Update docs about custom breadcrumb

covex-nn:
I am targeting this branch, because this is docs.

## Subject

Docs about custom breadcrumb was outdated.

### DIFF
--- a/docs/reference/breadcrumb.rst
+++ b/docs/reference/breadcrumb.rst
@@ -6,15 +6,16 @@ The ``SonataSeoBundle`` let's you create your own breadcrumbs based on your diff
 Create your own breadcrumb
 --------------------------
 
-First, you need to create a BlockService to handle your breadcrumbs. You can extend ``Sonata\SeoBundle\Block\BaseBreadcrumbMenuBlockService`` :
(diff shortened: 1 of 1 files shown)


docs/reference/breadcrumb.rst
```diff
@@ -23,16 +24,15 @@ First, you need to create a BlockService to handle your breadcrumbs. You can ext
          */
         public function getName()
         {
-            return 'acme.bundle.block.breadcrumb';
+            return 'app.block.breadcrumb';
```
kunicmarko20: shouldn't we return here the class name now?

covex-nn: I am not sure, but `Sonata\SeoBundle\Block\Breadcrumb\BaseBreadcrumbMenuBlockService::getName` returns `sprintf('Breadcrumb %s', $this->context)`. See [BaseBreadcrumbMenuBlockService.php#L69](https://github.com/sonata-project/SonataSeoBundle/blob/2.x/src/Block/Breadcrumb/BaseBreadcrumbMenuBlockService.php#L69)

Also `Sonata\BlockBundle\Block\BlockServiceInterface::getName` is used as a `$title` for `Sonata\CoreBundle\Model\Metadata`, so i guess `getName` can return anything.

jordisala1991:
Thank you @covex-nn

PR #274, opened 2018-06-15
Created SeoAwareTrait

core23:
I am targeting this branch, because this is a new feature.

## Changelog

```markdown
### Added
- Added `SeoAwareTrait`
```

## Subject

This trait can be used the same way as the `LoggerAwareTrait`, `ContainerAwareTrait`, ...

### DIFF
--- a/src/Seo/SeoAwareTrait.php
+++ b/src/Seo/SeoAwareTrait.php
@@ -1,0 +1,27 @@
+<?php
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\SeoBundle\Seo;
+
+trait SeoAwareTrait
+{
+    /**
+     * @var SeoPageInterface|null
+     *
+     * @required
+     */
+    protected $seoPage;
+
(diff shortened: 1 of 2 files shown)


src/Seo/SeoAwareTrait.php
```diff
@@ -0,0 +1,25 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\SeoBundle\Seo;
+
+trait SeoAwareTrait
+{
+    /**
+     * @var SeoPageInterface|null
```
kunicmarko20: What about adding `@required` it will try to autowire it.

core23: Didn't find any examples in the other traits.
Should you use autowiring when there is a setter method?

kunicmarko20: why not? You can still set it yourself or let it be autowired. A better question is, should we autowire it at all, I think it is not a best practice for 3rd party bundles.

OskarStark:
Do we need to tweak the docs or add a cookbook article about it?

kunicmarko20:
Yes, people need to know about it so they can use it.

src/Seo/SeoAwareTrait.php
core23: Fixed id

greg0ire:
Thanks @core23 !

PR #275, opened 2018-06-18
Require very last version of guzzle/guzzle

covex-nn:

## Subject

Trying to fix travis timeout with requiring very last version of `guzzle/guzzle`


greg0ire:
Thanks @covex-nn !

PR #276, opened 2018-07-05
Make alias public

core23:
I am targeting this branch, because this is a patch.

## Changelog

```markdown
### Fixed
- Make `sonata.seo.page` alias public
```

### DIFF
--- a/src/DependencyInjection/SonataSeoExtension.php
+++ b/src/DependencyInjection/SonataSeoExtension.php
@@ -67,7 +67,9 @@ class SonataSeoExtension extends Extension
         $definition->addMethodCall('setHtmlAttributes', [$config['head']]);
         $definition->addMethodCall('setSeparator', [$config['separator']]);
 
-        $container->setAlias('sonata.seo.page', $config['default']);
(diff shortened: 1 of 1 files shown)


src/DependencyInjection/SonataSeoExtension.php
```diff
@@ -67,7 +67,7 @@ protected function configureSeoPage(array $config, ContainerBuilder $container)
         $definition->addMethodCall('setHtmlAttributes', [$config['head']]);
         $definition->addMethodCall('setSeparator', [$config['separator']]);
 
-        $container->setAlias('sonata.seo.page', $config['default']);
+        $container->setAlias('sonata.seo.page', $config['default'])->setPublic(true);
```
kunicmarko20: This will fail

kunicmarko20: on < sf 3.3 because set alias wasn't returning alias. IIRC

kunicmarko20: yep, https://github.com/symfony/dependency-injection/blob/3.3/ContainerBuilder.php#L794

PR #281, opened 2018-10-02
Fixed possible null error in SeoPage

core23:
I am targeting this branch, because this is a patch.

Closes #{put_issue_number_here}

## Changelog

```markdown
### Fixed
- Initialized `SeoPage::$htmlAttributes` with empty array
```

## Subject

According to the PHPDoc, the value should always return an array.


greg0ire:
Thanks @core23 !

PR #266, opened 2018-02-18
Let page object also output a script tag to render json-ld data

ElectricMaxxx:
## Changelog

```markdown
### Added
- Added `StructuredDataAwarePage` to enable structured data support on seo pages
with an twig helper `{{ sonata_seo_structured_data() }}`
```

## Subject

In https://github.com/symfony-cmf/seo-bundle/issues/345 we want to introduce structured data. Doing so we do need a twig  helper and an extension for the Page class. That twig helper will render a script tag, where the whiches seem to communitate with each other.

PR #282, opened 2018-10-06
set explicit sphinx version

ElectricMaxxx:
To solve and use the correct behavior

## Changelog

Set explicit reuirements for docs now

### Changed
I user pip/python in an isolated environment. So i did here also, i created a new one and simply did a `pip install -r requirements.txt` and a `pip freeze > requirements.txt` this will set explicit versions of all used packages.
## Subject

To get PR #266 be green we need a newer/other sphinx version.


docs/Makefile
```diff
@@ -2,10 +2,10 @@
 #
 
 # You can set these variables from the command line.
-SPHINXOPTS    =
-SPHINXBUILD   = sphinx-build
-PAPER         =
-BUILDDIR      = _build
+SPHINXOPTS    :=
+SPHINXBUILD   := $(or ${SPHINXBUILD},${SPHINXBUILD},sphinx-build)
```
ElectricMaxxx: I can do a 
```bash
export SPHINXBUILD="~/.docs-ve/bin/sphinx-build -c /home/maximilian/OpenSource/SonataProject/SonataDoctrinePhpcrAdminBundle/docs/"
```
now and then run `make epub` or something else

OskarStark:
Hey Max, fist of all thank you for your work on this, but IIRC this change:

```diff
- Sphinx
+ Sphinx!=1.8.0
```

is enough an we don't need to handle the dependencies ourself.

//cc https://github.com/sonata-project/dev-kit/pull/424

ElectricMaxxx:
@OskarStark no worries, i simply whanted to inspect some stuff here, but we should add the change to use own builder in own environments

PR #284, opened 2018-10-10
DevKit updates for 2.x branch

SonataCI:

PR #285, opened 2018-10-10
Applied fixes from FlintCI

Soullivaneuh:


greg0ire:
Because of #284 I think

PR #288, opened 2018-11-02
Updated docs

core23:
I am targeting this branch, because this is related to the docs.

## Subject

Just a doc update

### DIFF
--- a/docs/reference/usage.rst
+++ b/docs/reference/usage.rst
@@ -9,26 +9,11 @@ By default, the object is initialized with values defined in the configuration:
 .. code-block:: yaml
 
     sonata_seo:
+        encoding: 'UTF-8'
         page:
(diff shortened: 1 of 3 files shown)


docs/reference/usage.rst
```diff
@@ -51,6 +36,19 @@ However, it is possible to alter these values at runtime:
         ->addMeta('property', 'og:description', $post->getAbstract())
     ;
 
+You could also prepend the page title, so that the global title is used as a suffix:
+
+.. code-block:: php
+
+    <?php
+
+    $seoPage = $this->container->get('sonata.seo.page');
```
OskarStark: should we mention DI instead of container get here?

core23: It's just an addition to the paragraph before

PR #322, opened 2019-01-20
Added support for latest exporter

core23:
## Subject

## Changelog

```markdown
### Added
- Added support for latest exporter
```

### DIFF
--- a/src/Sitemap/SourceManager.php
+++ b/src/Sitemap/SourceManager.php
@@ -13,8 +13,8 @@ declare(strict_types=1);
 
 namespace Sonata\SeoBundle\Sitemap;
 
-use Exporter\Source\ChainSourceIterator;
-use Exporter\Source\SourceIteratorInterface;
+use Sonata\Exporter\Source\ChainSourceIterator;
+use Sonata\Exporter\Source\SourceIteratorInterface;
 
 /**
  * Manager several chain source iterator grouped.
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
     ],
     "require": {
         "php": "^7.1",
-        "sonata-project/exporter": "^1.8",
+        "sonata-project/exporter": "^1.11 || ^2.0",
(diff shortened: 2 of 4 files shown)


src/Sitemap/SourceManager.php
```diff
@@ -13,8 +13,8 @@
 
 namespace Sonata\SeoBundle\Sitemap;
 
-use Exporter\Source\ChainSourceIterator;
-use Exporter\Source\SourceIteratorInterface;
+use Sonata\Exporter\Source\ChainSourceIterator;
+use Sonata\Exporter\Source\SourceIteratorInterface;
```
greg0ire: Please trigger autoload for that one (it is used in `addSource`'s signature)

core23: I'm not sure if this is the right solution...

You didn't add the trigger neither
https://github.com/sonata-project/SonataAdminBundle/pull/5408

greg0ire: In that PR, the type hint appear in a constructor, and constructors don't have to be compatible with the constructor of the parent class. If you want to be sure this is the right solution, let me provide you with a test.

greg0ire:
```
use Exporter\Source\SourceIteratorInterface;
use Sonata\SeoBundle\Sitemap\SourceManager;

final class ExtendingSourceManager extends SourceManager
{
    public function addSource($group, SourceIteratorInterface $source, array $types = [])
    {
    }
}
```
You can test that this crashes, and that it no longer crashes with the `class_exists`.

greg0ire: Well… I feel dumb now, this does not crash.

greg0ire: Ah for that one I know, `SourceManager` implements `SourceManagerInterface`, and `implements` triggers autoloading :) . That one makes sense 👍

greg0ire:
Thanks @core23 !

PR #323, opened 2019-01-23
DevKit updates for 2.x branch

SonataCI:


OskarStark:
@greg0ire please review my commits

tests/DependencyInjection/ConfigurationTest.php
```diff
@@ -22,11 +22,22 @@ class ConfigurationTest extends TestCase
...
+                'head' => [],
+                'metas' => [],
+                'separator' => ' - ',
+                'title' => 'Project name',
+            ],
+            'sitemap' => [
+                'doctrine_orm' => [],
+                'services' => [],
+            ],
+        ];
```
greg0ire: So, that's because the order is not the same, right… not great, but not sure there is a better solution…

OskarStark: exactly

OskarStark: but would fit our needs right now

PR #325, opened 2019-01-23
Added the SeoPageInterface to the container as an alias.

### LINKED ISSUE #320: Make an alias Service So the SeoPageInterface can be autowired
## Feature Request

Right now whenever you want to alter the Seo attributes, you need to do the following:

```
$seoPage = $this->get('sonata.seo.page')
```

This is only possible when you have the `Container` available. That's not always the case. 
Since Symfony uses `autowiring` and `autoconfiguring` it would be great to just use `DependencyInjection`. This means that developers just use the `SeoPageInterface` whenever they want. 

I am aware you can configure every service to use the `@sonata.seo.page` service, but that will end up in a lot of configuration while it can be reduced to a Service alias.

leroy0211:
## Subject

Added the `SeoPageInterface` as an alias in the container. This way developers can use Dependency Injection to grab the SeoPage class they configured in their configuration. 

The `SeoPageInterface` wil always be the instance which is configured in the `sonata_seo.page.default` setting.

Closes #320 

## Changelog

```markdown
### Added
- Added `SeoPageInterface` as alias service of `sonata.seo.page` for Dependency Injection usages.
```

### DIFF
--- a/src/DependencyInjection/Compiler/ServiceCompilerPass.php
+++ b/src/DependencyInjection/Compiler/ServiceCompilerPass.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace Sonata\SeoBundle\DependencyInjection\Compiler;
 
+use Sonata\SeoBundle\Seo\SeoPageInterface;
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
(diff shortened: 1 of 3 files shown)


OskarStark:
Thank you, I will ask for a new release internally

PR #327, opened 2019-01-26
Added title prefix and suffix options with configuration

### LINKED ISSUE #157: Simple way to implement prefix (prepend) or sufix (appened) in titles
I know how to add it from controller, but I was wondering if there is easy way adding it via configuration, rather then controller. 

``` yaml
sonata_seo:
    page:
        title: "%%translationKey%% - Company Name"
```

so translation key can be replaced by any added title.

### LINKED ISSUE #319: Add configurable prefix and suffix for the <title>
## Feature Request

It would be great to have a configurable prefix and suffix for the `<title>` element. 

When you use `->setTitle()` or `->addTitle()` the prefix and suffix will be added to the result. The prefix and suffix should be configured in the configuration but may be overwritten by the user when requesting the `sonata.seo.page` service.

leroy0211:
## Subject

The title prefix and/or suffix can be configured and altered on runtime. The default title is moved to the default suffix. This way when setting a new title, the prefix and suffix wil be kept intact.

The Bundle configuration `page.title` is changed to support a default prefix and suffix, but BC is kept intact!

Closes #319 
Closes #157

## Changelog

```markdown
### Added
- Added configurable page title pre- and suffix

### Changed
- `sonata_seo.page.title` section extended with `prefix` and `suffix`, including BC fix.
```

    
## To do

- [ ] Update the documentation with tagging information for prior symfony 3.3 projects


leroy0211:
Update on the 2 issues found by Scrutinizer.

It has got to do with the PHP7 notation of returning `self` in the `SeoPageInterface`. I've looked up how to handle fluent interfaces, and it just should return the `SeoPageInterface`. But after running the `php-cs-fixer` it got replaced by `self`. I'm not sure this is a real issue or not. 

If you want me to remove the PHP7 notation, I'll do that instead.

greg0ire:
It's the same thing, I don't see any issue here 👍

docs/reference/usage.rst
```diff
@@ -34,17 +34,81 @@ However, it is possible to alter these values at runtime::
...
-You could also prepend the page title, so that the global title is used as a suffix::
+.. note::
+
+    If you need a shortcut to inject the seo page into any class, you could use the
+    ``Sonata\SeoBundle\Seo\SeoAwareTrait``.
+
+Pre- or suffix the page title
+-------------------------------
+
+If you want to change the suffix of your page titles, you can either do this by altering the ``title`` directly or use the ``title.suffix``.
```
greg0ire: Please comply with the sf docs standards:

> [Each line should break approximately after the first word that crosses the 72nd character (so most lines end up being 72-78 characters);](http://symfony.com/doc/current/contributing/documentation/standards.html)

docs/reference/usage.rst
```diff
@@ -34,17 +34,81 @@ However, it is possible to alter these values at runtime::
         ->addMeta('property', 'og:description', $post->getAbstract())
     ;
 
-You could also prepend the page title, so that the global title is used as a suffix::
+.. note::
+
+    If you need a shortcut to inject the seo page into any class, you could use the
+    ``Sonata\SeoBundle\Seo\SeoAwareTrait``.
+
+Pre- or suffix the page title
```
greg0ire:
```suggestion
Prefix or suffix the page title
```

docs/reference/usage.rst
```diff
@@ -34,17 +34,81 @@ However, it is possible to alter these values at runtime::
...
+-------------------------------
```
greg0ire:
```suggestion
---------------------------------
```

docs/reference/usage.rst
```diff
@@ -34,17 +34,81 @@ However, it is possible to alter these values at runtime::
...
+
+    sonata_seo:
+        page:
+            title:
+                prefix: 'My Prefix'
+                suffix: 'My Suffix'
+
+You can also edit the pre- or suffix at runtime, however this is uncommon::
+
+    $seoPage = $this->container->get('sonata.seo.page');
```
greg0ire: Getting services from the service container is a bad practice and should not be encouraged. Please recommend using dependency injection instead.

src/Seo/SeoPage.php
```diff
@@ -23,6 +23,16 @@ class SeoPage implements SeoPageInterface
      */
     protected $title;
 
+    /**
+     * @var string
+     */
+    protected $titlePrefix;
```
```suggestion
    private $titlePrefix;
```

src/Seo/SeoPage.php
```diff
@@ -23,6 +23,16 @@ class SeoPage implements SeoPageInterface
...
+
+    /**
+     * @var string
+     */
+    protected $titleSuffix;
```
```suggestion
    private $titleSuffix;
```

src/Seo/SeoPage.php
```diff
@@ -105,7 +115,21 @@ public function addTitle($title)
      */
     public function getTitle()
     {
-        return $this->title;
+        return implode($this->separator, array_filter([$this->titlePrefix, $this->title, $this->titleSuffix]));
+    }
+
+    public function setTitlePrefix(string $prefix): SeoPageInterface
```
```suggestion
    final public function setTitlePrefix(string $prefix): SeoPageInterface
```

src/Seo/SeoPage.php
```diff
@@ -105,7 +115,21 @@ public function addTitle($title)
...
+    {
+        $this->titlePrefix = $prefix;
+
+        return $this;
+    }
+
+    public function setTitleSuffix(string $suffix): SeoPageInterface
```
```suggestion
    final public function setTitleSuffix(string $suffix): SeoPageInterface
```

src/Seo/SeoPageInterface.php
```diff
@@ -34,6 +34,20 @@ public function addTitle($title);
      */
     public function getTitle();
 
+    /**
+     * @param string $prefix
+     *
+     * @return SeoPageInterface
+     */
```
greg0ire: Please remove both docblocks, they don't add any value.

src/Seo/SeoPageInterface.php
```diff
@@ -34,6 +34,20 @@ public function addTitle($title);
...
+    public function setTitlePrefix(string $prefix): self;
```
greg0ire: Also, adding methods to an interface is a BC-break, it should be a separate interface.

docs/reference/usage.rst
leroy0211: Only on my added documentation? Because some lines which already exists crosses the 72 characters.

docs/reference/usage.rst
leroy0211: Should I update the other documentation too with a Controller example?

src/Seo/SeoPageInterface.php
leroy0211: I agree, but if someone updates to a new minor, they manually have to update their composer file, if they done it right. SonataSeoBundle also dropped PHP5 support in a minor, isn't that considered a BC break too?
I gladly would make this an interface with a trait combination, but your feedback contradicts what others have said in PR #325

leroy0211: I was referring to the part in which PHP5 was dropped. But yes, you are right. composer won't let you install that version. 

I taught by adding a trait for a new interface would be useful, because it's just a few setters but without any getters a trait isn't that useful in this case.

docs/reference/usage.rst
leroy0211: Does this also mean that `SeoAwareTrait` is deprecated, now can use dependency injection? If so, which version should I add in the deprecation message?

greg0ire: Link for the lazy: https://github.com/sonata-project/SonataSeoBundle/blob/2.x/src/Seo/SeoAwareTrait.php

greg0ire: And I think that trait could indeed be deprecated in favor of constructor injection.

docs/reference/usage.rst
```diff
@@ -17,34 +14,144 @@ By default, the object is initialized with values defined in the configuration:
...
+defined in ``sonata_seo.page.default``.
+
+.. code-block:: php
+
+    // src/Controller/PostController.php
+    namespace App\Controller;
+
+    use Sonata\SeoBundle\Seo\SeoPageInterface;
+
+    class PostController
```
greg0ire:
```suggestion
    final class PostController
```

docs/reference/usage.rst
```diff
@@ -17,34 +14,144 @@ By default, the object is initialized with values defined in the configuration:
...
+    {
+        private $seoPage;
+
+        public function __construct(SeoPageInterface $seoPage)
+        {
+            $this->seoPage = $seoPage;
+        }
+
+        public function index()
```
greg0ire:
```suggestion
        public function index(): Response
```

docs/reference/usage.rst
```diff
@@ -17,34 +14,144 @@ By default, the object is initialized with values defined in the configuration:
...
+You can alter the SEO information at runtime by injecting the ``SeoPageInterface``
+into your classes. The bundle wil automatically use the configured service
+defined in ``sonata_seo.page.default``.
+
+.. code-block:: php
+
+    // src/Controller/PostController.php
+    namespace App\Controller;
+
+    use Sonata\SeoBundle\Seo\SeoPageInterface;
```
greg0ire:
```suggestion
    use Sonata\SeoBundle\Seo\SeoPageInterface;
    use Symfony\Component\HttpFoundation\Response;
```

docs/reference/usage.rst
```diff
@@ -17,34 +14,145 @@ By default, the object is initialized with values defined in the configuration:
...
+    final class PostController
+    {
+        private $seoPage;
+
+        public function __construct(SeoPageInterface $seoPage)
+        {
+            $this->seoPage = $seoPage;
+        }
+
+        public function index(): Response
```
OskarStark: Shouldnt we inject it in the action directly?

greg0ire: :thinking: why? Is it computed from the request?

src/Seo/SeoAwareTrait.php
```diff
@@ -13,6 +13,11 @@
 
 namespace Sonata\SeoBundle\Seo;
 
+@trigger_error('The '.__NAMESPACE__.'\SeoAwareTrait trait deprecated since version 2.8 and will be removed in 3.0. Inject the Sonata\SeoBundle\Seo\SeoPageInterface in your classes instead.', E_USER_DEPRECATED);
```
greg0ire:
```suggestion
@trigger_error(
    'The '.__NAMESPACE__.'\SeoAwareTrait trait deprecated since version 2.8 and will be removed in 3.0. '.
    'Inject the Sonata\SeoBundle\Seo\SeoPageInterface in your classes instead.', 
    E_USER_DEPRECATED
);
```

src/Seo/SeoPageInterface.php
```diff
@@ -34,6 +34,10 @@ public function addTitle($title);
      */
     public function getTitle();
 
+    public function setTitlePrefix(string $prefix): self;
+
+    public function setTitleSuffix(string $suffix): self;
```
greg0ire: So let's think of another name for the new interface. It describes the interface of a page whose title can be suffixed or prefixed. `PageWithAffixableTitleInterface`?

leroy0211: I was thinking of adding all `title` logic (`setTitle`, `getTitle` `addTitle`, `setTitlePrefix` and `setTitleSuffix`) to this new interface and call it `SeoPageTitleInterface`.  In a next major those methods can be removed from the `SeoPageInterface`. Combining interfaces with the same method definition isn't going to break.
What do you think?

greg0ire: That sounds great, please do that!

src/DependencyInjection/Compiler/ServiceCompilerPass.php
```diff
@@ -28,8 +30,14 @@ public function process(ContainerBuilder $container)
 
         $definition = $container->findDefinition($config['default']);
 
-        $definition->addMethodCall('setTitlePrefix', [$config['title']['prefix']]);
-        $definition->addMethodCall('setTitleSuffix', [$config['title']['suffix']]);
+        $definitionClass = $this->getDefinitionClassname($container, $definition);
+        if (is_subclass_of($definitionClass, SeoTitleInterface::class)) {
```
greg0ire: Why not use `instanceof` here?

greg0ire: Ah cause it's a string, alright!

src/DependencyInjection/Compiler/ServiceCompilerPass.php
```diff
@@ -40,4 +48,21 @@ public function process(ContainerBuilder $container)
...
+        }
+
+        if (!$container->hasParameter(trim($definitionClass, '%'))) {
+            return null;
+        }
+
+        $definition = new Definition($container->getParameter(trim($definitionClass, '%')));
+
+        return $this->getDefinitionClassname($container, $definition);
+    }
```
greg0ire: I think we could bump Symfony from 2.8 to 3.4 in composer.json (we were waiting for a good reason to do that), and then use [autoconfiguration](https://symfony.com/blog/new-in-symfony-3-3-service-autoconfiguration) to tag all instances of the new interface with a specific tag. Then we could make the calls based on that tag and get rid of this method.

src/Seo/SeoTitleInterface.php
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+namespace Sonata\SeoBundle\Seo;
+
+interface SeoTitleInterface
+{
+    /**
+     * @param string $title
+     *
+     * @return SeoPageInterface
```
greg0ire: Not sure what to think about this… maybe we should change it too? Also, shoud `SeoTitleInterface` extend `SeoPageInterface`?

src/DependencyInjection/Compiler/ServiceCompilerPass.php
```diff
@@ -28,8 +30,14 @@ public function process(ContainerBuilder $container)
...
         $definition = $container->findDefinition($config['default']);
 
-        $definition->addMethodCall('setTitlePrefix', [$config['title']['prefix']]);
-        $definition->addMethodCall('setTitleSuffix', [$config['title']['suffix']]);
+        $definitionClass = $this->getDefinitionClassname($container, $definition);
+        if (is_subclass_of($definitionClass, SeoTitleInterface::class)) {
+            $definition->addMethodCall('setTitlePrefix', [$config['title']['prefix']]);
+            $definition->addMethodCall('setTitleSuffix', [$config['title']['suffix']]);
+        } else {
+            $definition->addMethodCall('setTitle', [$config['title']['suffix']]);
```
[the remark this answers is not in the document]
leroy0211: It's for backward compatibility of functionality which previously existed. If someone did not implement the `SeoTitleInterface` in their own implementation of `SeoPageInterface`, the title wil be set.

src/DependencyInjection/Compiler/ServiceCompilerPass.php
leroy0211: I'm super fan of the auto-tagging feature, but I'm not sure how this would work in this case. Because then you make title, html/head/body attributes, etc. all  based on string properties like `$seo->get('title');` and `$seo->get('meta')`. which in turn return either a string, parameterbag, etc. 

Or what do you exactly mean by this?

src/DependencyInjection/Compiler/ServiceCompilerPass.php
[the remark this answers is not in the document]
greg0ire: Yes, but… it will be set to a suffix?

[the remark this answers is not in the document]
leroy0211: Yes because in the previous documentation you could use `addTitle()` to prepend the title you configured in the config. So that would bring that functionality back. Otherwise I have to remove it, but then the `sonata_seo.page.title` option doesn't do anything.

I had to do this because the suffix needed to be moved to another interface.

src/DependencyInjection/Compiler/ServiceCompilerPass.php
greg0ire: I'm not following you, what does this have to do with html/head/body attributes, etc. all based on string properties like $seo->get('title'); ?

I mean that we could auto-tag instance of SeoTitleInterface with a special tag.
Then, in a compiler pass, maybe this one, maybe a new one, we would find services with that tag, and then add method calls to `setTitleSuffix` and `setTitlePrefix` on them. That way you should not need `getDefinitionClassname` at all.

src/Seo/SeoTitleInterface.php
leroy0211: Maybe. But what would you do when adding a new feature again? Would you then extend the `SeoTitleInterface`, so you get an extend chain of interfaces adding features over and over?

src/DependencyInjection/Compiler/ServiceCompilerPass.php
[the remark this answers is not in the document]
greg0ire: Oh ok, great! Tiny thing: you said "prepend", did you mean "append"? Because then the prefix should be used, right?

src/Seo/SeoTitleInterface.php
greg0ire: Hmmm good point that does not sound right… maybe we should just return `static` instead of `self`?

src/DependencyInjection/Compiler/ServiceCompilerPass.php
leroy0211: Oh, then I misunderstood. That's not what that method does.

The `getDefinitionClassname` method is used because you can only grab the class property defined in the definition. You define the classname in the definition directly, or you use a parameter like this bundle does: `sonata.seo.page.default.class` in `services.xml`. 

This method just tries to resolve both so I can test which method-call needs to be passed.

As example, I have added a yaml notation of what I mean.
```
# services.yml
parameters:
    sonata.seo.page.default.class: 'Sonata\SeoBundle\Seo\SeoPage'

sonata.seo.page.default:
    class: '%sonata.seo.page.default.class%'
```

When getting the `sonata.seo.page.default` definition the parameters are not yet resolved, so using `$definition->getClass()` will just return the string `%sonata.seo.page.default.class%`.

And because the class does not exist, it will try to resolve that paramater from the container by trimming the `percentages`. That's all that method does. 

You can not get the service from the container yet, because it's still compiling, because it's a CompilerPass.

src/DependencyInjection/Compiler/ServiceCompilerPass.php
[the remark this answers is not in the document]
leroy0211: No, prepend.

Copy from the original documentation:

> You could also prepend the page title, so that the global title is used as a suffix

So without breaking BC I've added the `setTitle` as methodcall, because that's what was [previously done too](https://github.com/sonata-project/SonataSeoBundle/pull/327/files#diff-1c4ead08f7be66cf6552316a451c0c42L31).

src/Seo/SeoTitleInterface.php
leroy0211: You can not use the static as return type declaration in PHP 7.

src/DependencyInjection/Compiler/ServiceCompilerPass.php
greg0ire: > I only added this interface because modifying the SeoPageInterface is a BC break.

That was already clear

> You have only 1 active instance of SeoPageInterface in your entire application, right?

That too, but I don't know another way to detect an interface is implemented by a service (and the leverage Symfony instead of using `getDefinitionClassname` than using `_instanceof`. Not that there is an equivalent for php though: https://github.com/symfony/symfony-docs/blob/4b549cc81a46113ec582e765f7f05053431aa2ec/service_container/tags.rst#autoconfiguring-tags

greg0ire: Here is where the magic happens: https://github.com/symfony/symfony/blob/5aa0967f9f0ab803ededefb040d48a0ebc7a27a6/src/Symfony/Component/DependencyInjection/Compiler/ResolveInstanceofConditionalsPass.php#L68

src/Seo/SeoTitleInterface.php
greg0ire: I think we should use `self` here. Returning what is potentially another object is not fluent, and we are aiming at fluent here.

src/DependencyInjection/Compiler/ServiceCompilerPass.php
greg0ire: And here you have parameter resolution: https://github.com/symfony/symfony/blob/5aa0967f9f0ab803ededefb040d48a0ebc7a27a6/src/Symfony/Component/DependencyInjection/Compiler/ResolveInstanceofConditionalsPass.php#L55

leroy0211: > but I don't know another way to detect an interface is implemented by a service (and the leverage Symfony instead of using getDefinitionClassname than using _instanceof.

So you suggest to use service tagging just to test if a service has implemented a specific interface? Isn't that "abusing" the tagging system for something else?

greg0ire: Yes, that's exactly what I'm suggesting, and I would call that reusing code that is thoroughly tested, in ordered to achieve something it was designed for. In our case, there is only one implementation, but still… and we don't actually know if the person reimplementing `sonata.seo.page` does not implement it as a wrapper for several classes, all implementing `SeoTitleInterface`. There could very well be several implementations, we do not know.

leroy0211: I've updated it so it uses the tag-system. Could you take a look?

tests/DependencyInjection/Compiler/AlterAliasPublicationPass.php
```diff
@@ -0,0 +1,52 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\SeoBundle\Tests\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+class AlterAliasPublicationPass implements CompilerPassInterface
```
greg0ire:
```suggestion
final class AlterAliasPublicationPass implements CompilerPassInterface
```

tests/DependencyInjection/Compiler/AlterAliasPublicationPass.php
```diff
@@ -0,0 +1,52 @@
...
+     * @var string
+     */
+    private $service;
+
+    public function __construct(string $service)
+    {
+        $this->service = $service;
+    }
+
+    public function process(ContainerBuilder $container)
```
greg0ire:
```suggestion
    public function process(ContainerBuilder $container): void
```

tests/DependencyInjection/Compiler/AlterAliasPublicationPass.php
```diff
@@ -0,0 +1,52 @@
...
+    {
+        if (null === $definition = $this->getDefinitionOrAlias($container)) {
+            return;
+        }
+
+        $definition->setPublic(true);
+    }
+
+    private function getDefinitionOrAlias(ContainerBuilder $container)
```
greg0ire:
```suggestion
    /**
     * @return Definition|Alias
     */
    private function getDefinitionOrAlias(ContainerBuilder $container)
```

greg0ire: Please also add use statements for both things

src/DependencyInjection/SonataSeoExtension.php
```diff
@@ -31,27 +32,32 @@ class SonataSeoExtension extends Extension
...
-        $loader->load('commands.xml');
-
         $this->configureSeoPage($config['page'], $container);
         $this->configureSitemap($config['sitemap'], $container);
 
         $container->getDefinition('sonata.seo.twig.extension')
             ->replaceArgument(1, $config['encoding']);
+
+        // Register autoconfiguration rules for Symfony DI 3.3+
+        if (method_exists($container, 'registerForAutoconfiguration')) {
```
greg0ire: I think you can bump the requirements of this bundle to Symfony 3.4 and get rid of that condition.

tests/DependencyInjection/Compiler/AlterAliasPublicationPass.php
greg0ire: fixed

tests/Seo/SeoPageTest.php
```diff
@@ -119,6 +119,27 @@ public function testAddTitle()
...
+
+        $page->setTitleSuffix('My Suffix');
+
+        $this->assertEquals('My Prefix - Additional title - My title - My Suffix', $page->getTitle());
+
+        $page->setTitle('');
+
+        $this->assertEquals('My Prefix - My Suffix', $page->getTitle());
+
+        $page->setTitlePrefix('');
```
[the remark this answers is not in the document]
leroy0211: Missed that assertion.

docs/reference/usage.rst
```diff
@@ -17,34 +14,145 @@ By default, the object is initialized with values defined in the configuration:
...
+
+    // src/Controller/PostController.php
+    namespace App\Controller;
+
+    use Sonata\SeoBundle\Seo\SeoPageInterface;
+    use Symfony\Component\HttpFoundation\Response;
+
+    final class PostController
+    {
+        private $seoPage;
```
core23: You could use the SeoAwareTrait here

OskarStark: No it was deprecated in this PR

PR #328, opened 2019-01-27
Added body attributes including a new twig helper

### LINKED ISSUE #317: Add support for body attributes
## Feature Request

This bundle comes packages with `sonata_seo_head_attributes()` as twig function, but it would be great to also add `sonata_seo_body_attributes()` to be more complete.

leroy0211:
## Subject

This PR will add the body attributes feature so body attributes can be altered using SeoBundle too. Including a twig helper function which can be used. 

Closes #317 

## Changelog

```markdown
### Added
- Added body attributes functionality 
- Added twig function to render the body attributes

```

    
## To do

- [ ] Update the documentation


src/Seo/BodyAttributeInterface.php
```diff
@@ -0,0 +1,27 @@
...
+
+interface BodyAttributeInterface
+{
+    public function setBodyAttributes(array $attributes): self;
+
+    public function addBodyAttribute(string $name, string $value): self;
+
+    public function removeBodyAttribute(string $name): self;
+
+    public function getBodyAttributes(): array;
```
greg0ire: Please document the return type hint (@return `string[]` ?)

src/Seo/BodyAttributeInterface.php
```diff
@@ -0,0 +1,27 @@
...
+
+namespace Sonata\SeoBundle\Seo;
+
+interface BodyAttributeInterface
+{
+    public function setBodyAttributes(array $attributes): self;
+
+    public function addBodyAttribute(string $name, string $value): self;
+
+    public function removeBodyAttribute(string $name): self;
```
greg0ire: I don't think these 3 methods should be part of the interface: if someone wants to add their own immutable implementation, they should be able to use a constructor to provide the attributes instead. Also, if someone want to create a cached implementation, they should not have to reimplement these methods. And finally, you do not rely on these methods in the TwigExtension

src/Seo/BodyAttributeInterface.php
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\SeoBundle\Seo;
+
+interface BodyAttributeInterface
```
greg0ire: You use `BodyAttribute` here, and `BodyAttributes` in the trait… how about switching to `BodyAttributeBag`?

src/Seo/BodyAttributesTrait.php
```diff
@@ -0,0 +1,50 @@
...
+    public function getBodyAttributes(): array
+    {
+        return $this->bodyAttributes;
+    }
+
+    public function hasBodyAttribute(string $name): bool
+    {
+        return isset($this->bodyAttributes[$name]);
+    }
+}
```
greg0ire: All these methods should be `final`

src/Seo/SeoPage.php
```diff
@@ -16,8 +16,10 @@
 /**
  * http://en.wikipedia.org/wiki/Meta_element.
  */
-class SeoPage implements SeoPageInterface
+class SeoPage implements SeoPageInterface, BodyAttributeInterface
 {
+    use BodyAttributesTrait;
```
greg0ire: Consider creating a new class instead… I mean there are `html` attributes, `head` attributes, and now `body` attributes. IMO we should migrate to 3 object properties, that would implement `Traversable`

src/Twig/Extension/SeoExtension.php
```diff
@@ -187,6 +189,20 @@ public function getHeadAttributes()
         return rtrim($attributes);
     }
 
+    public function getBodyAttributes(): string
```
greg0ire: How about returning a hash and providing a twig template that would do this instead? This is Twig's job.

leroy0211: All these changes are in compliance with the already chosen implementation in `SeoPage`. Do you suggest to migrate all implementation keeping BC? Because I can change that, but maybe that's better for a new major version?

greg0ire: I do suggest to migrate all implementations keeping BC, but not necessarily in this PR. New major versions should only drop BC-hacks.

src/Seo/BodyAttributeInterface.php
leroy0211: All methods (both getters and setters) should be available in the `SeoPageInterface`, because the main functionality is to alter them at runtime. And because you can configure your own SeoPage instance in `sonata_seo.page.default` this HAS to contain all the methods. But I can not extend the SeoPageInterface with other interfaces, because that would break BC. 

The other option is to remove the `sonata_seo.page.default` option from the configuration, so you can not use your own service implementation of `SeoPageInterface`, and Sonata Seo wil always use it's own `SeoPage` class. But then you don't actually need the `SeoPageInterface` anymore.

leroy0211: But then you are forcing people to write their own implementation of `SeoPage` if they want to alter data (like the html/head/body attributes). So if people must write their own implementation of `SeoPage`, they can drop this bundle at all. Because that IS the power of this bundle! Altering Seo data on different kind of pages. 

For example on the homepage, I want to display `<body id="homepage">`, so my seo javascript (eg: google tag manager) can act on this page. In my `HomepageController` I just need to add `$seoPage->addBodyAttribute('id', 'homepage')`

Just like you would add html or head attributes on runtime, or a title, title-prefix or title-suffix like in the other PR (#327)

greg0ire: Oh I see, you're right, the methods should probably stay!

Here is how it could look though: `$seoPage->bodyTagAttributes()->add('id', 'homepage');`. That way `SeoPage` stays not too big, and we enforce a consistent interface for every group of attributes.

leroy0211: Ok, let me see if I can change that like that.

greg0ire: Thank you !

leroy0211: I've changed this. I still got a question about this. Some of the methods are now deprecated, but are still used in the tests, this wil give some deprecation notices. Should those be resolved by removing the deprecated tests? Or just leave them in place?

greg0ire: Leave them in place, but add them to the `legacy` group, by adding the `@group legacy` annotation to their phpdocs

leroy0211: Ooh great! Didn't know that.

src/Seo/BodyAttributeInterface.php
leroy0211: I have removed both the interface and trait as you suggested.

src/Seo/AttributeBag.php
```diff
@@ -0,0 +1,62 @@
...
+ */
+
+namespace Sonata\SeoBundle\Seo;
+
+class AttributeBag implements \IteratorAggregate
+{
+    private $attributes = [];
+
+    /**
+     * @param array $attributes
```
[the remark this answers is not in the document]
leroy0211: I just accepted the suggestion, but it's kind of mixed actually. Because you can use strings/booleans/etc. However I think based on the feature it covers (html/head/body attributes) it's quite uncommon to NOT have string attributes.

src/Resources/views/attributes.html.twig
```diff
@@ -0,0 +1,8 @@
+{%- for attrname, attrvalue in attr -%}
+    {{- " " -}}
+    {%- if attrvalue is same as(true) -%}
+        {{- attrname }}="{{ attrname }}"
+    {%- elseif attrvalue is not same as(false) -%}
+        {{- attrname }}="{{ attrvalue }}"
+    {%- endif -%}
+{%- endfor -%}
```
greg0ire: Please configure your IDE to always add a newline at end of file, this will make the red pictogram disappear.

src/Seo/SeoPage.php
```diff
@@ -72,8 +77,9 @@ public function __construct($title = '')
             'property' => [],
         ];
 
-        $this->htmlAttributes = [];
-        $this->headAttributes = [];
+        $this->htmlAttributes = new AttributeBag();
+        $this->headAttributes = new AttributeBag();
```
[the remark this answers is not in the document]
leroy0211: What should I do with the adders and setters? Pass the value only to the `AttributeBag` like I do now? Or to both `AttributeBag` and the protected properties?

src/Resources/views/attributes.html.twig
leroy0211: Ok, didn't know it was common practice to also add a new blank end line to twig files.

src/Seo/SeoPage.php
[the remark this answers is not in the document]
greg0ire: To both, I think… extending classes expect these properties to be filled. This is why we should use `private` and `final` as much as we can :sweat_smile: … refactorings are a lot easier when you do.

src/Seo/SeoPage.php
```diff
@@ -213,72 +221,88 @@ public function removeHtmlAttributes($name)
...
-     * @return array
+     * @return bool
      */
     public function hasHeadAttribute($name)
     {
-        return isset($this->headAttributes[$name]);
+        return $this->headAttributes()->has($name);
+    }
+
+    public function htmlAttributes(): AttributeBag
```
```suggestion
    final public function htmlAttributes(): AttributeBag
```

src/Seo/SeoPage.php
```diff
@@ -213,72 +221,88 @@ public function removeHtmlAttributes($name)
...
+    {
+        return $this->htmlAttributes;
+    }
+
+    public function headAttributes(): AttributeBag
```
```suggestion
    final public function headAttributes(): AttributeBag
```

src/Seo/SeoPage.php
```diff
@@ -213,72 +221,88 @@ public function removeHtmlAttributes($name)
...
+    {
+        return $this->headAttributes;
+    }
+
+    public function bodyAttributes(): AttributeBag
```
```suggestion
    final public function bodyAttributes(): AttributeBag
```

src/Twig/Extension/SeoExtension.php
```diff
@@ -174,17 +175,20 @@ public function renderHeadAttributes()
...
+            $attributes = $this->page->headAttributes();
+        } else {
+            $attributes = new AttributeBag($this->page->getHeadAttributes());
         }
 
-        return rtrim($attributes);
+        return $this->renderAttributes($attributes, $environment);
+    }
+
+    public function getBodyAttributes(\Twig_Environment $environment): string
```
greg0ire:
```suggestion
    final public function getBodyAttributes(\Twig_Environment $environment): string
```

src/Seo/SeoPage.php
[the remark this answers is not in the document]
leroy0211: Yes. I am fully aware of that by now! 😆

src/Twig/Extension/SeoExtension.php
```diff
@@ -264,4 +268,17 @@ private function normalize($string)
     {
         return htmlentities(strip_tags($string), ENT_COMPAT, $this->encoding);
     }
+
+    private function renderAttributes(AttributeBag $attributes, \Twig_Environment $environment = null): string
```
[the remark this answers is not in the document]
leroy0211: I didn't used them because the Extension itself is extended from `\Twig_Extension`, just to keep it common. But I shall update that one too.

src/Twig/Extension/SeoExtension.php
```diff
@@ -264,4 +268,17 @@ private function normalize($string)
...
+    {
+        if (null === $environment) {
+            return '';
+        }
+
+        try {
```
leroy0211: What do you tink. Should I use this try catch for the template rendering? Or just let the exception bubble up?

src/Seo/AttributeBag.php
```diff
@@ -0,0 +1,62 @@
...
+    {
+        $this->attributes = $attributes;
+    }
+
+    public function add(string $name, $value): void
+    {
+        $this->attributes[$name] = $value;
+    }
+
+    public function get(string $name, $default = null): ?string
```
core23: `string` is wrong here, because you could add any type using the `add` method

src/Twig/Extension/SeoExtension.php
greg0ire: I think you should let the exception bubble up, let people know where they failed.

tests/Twig/Extension/SeoExtensionTest.php
```diff
@@ -146,4 +192,12 @@ public function testOEmbedLinks()
             $extension->getOembedLinks()
         );
     }
+
+    protected function getEnvironment()
+    {
+        $loader = new FilesystemLoader();
+        $loader->addPath(__DIR__.'/../../../src/Resources/views', 'SonataSeo');
+
+        return new \Twig_Environment($loader);
```
greg0ire: Needs to be changed too please

src/Seo/AttributeBag.php
greg0ire: Could you? in the end, this appears in an html attribute, for instance `data-count="2.00"`. For html, everything is a string, so it's probably better to force string everywhere so that people make sure to handle the `float` to `string` conversion themselves.

OskarStark: In Riga case we Must type gibt the value if the add method above with string too.

I think @core23 is talking about the value, right?

greg0ire: Uuuuh wie bitte?

OskarStark: Sorry, I updated my comment 😅

OskarStark: Sorry, I updated my comment 😅

src/Seo/SeoPage.php
```diff
@@ -183,6 +205,8 @@ public function setHtmlAttributes(array $attributes)
     {
         $this->htmlAttributes = $attributes;
 
+        $this->htmlAttributes()->set($attributes);
```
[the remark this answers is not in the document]
greg0ire: Nope, because that's an array

src/Seo/AttributeBag.php
leroy0211: > Yup so… any good reason to use a non-string value here?

If it's all based on html attributes then a string will be correct. However some attributes have a boolean state, which can either be formatted as `<elem disabled />` or `<elem disabled="anystring|true|false" />` 
A boolean attribute wil only check for the presence of the attribute. If you use `attr="false"` as syntax, it is considered truethfull. 

So if you want eg. to add the `disabled` attribute, you have to write it like `->add('disabled', 'disabled')`. Which in other libraries is sometimes also written like `->add('disabled', true)` or `->add('disabled', false)` (to remove the attribute).

tests/Twig/Extension/SeoExtensionTest.php
leroy0211: oops. missed that.

src/Seo/AttributeBag.php
greg0ire: Could that somehow be useful in the context of SEO?

leroy0211: I'm not sure, because every tracker-feature requires something different. Taking a quick look at the [available meta's](https://gist.github.com/kevinSuttle/1997924) and the [opengraph protocol](http://opengraphprotocol.org/) every attribute has a value.

So forcing it to use a string should be valid. If there is a use-case in which it's a boolean, for example in the opengraph protocol, it should be either `"true"`, `"1"` or `"yes"` for truthful and  `"false"`, `"0"` or `"no"` for falsified.

greg0ire: The class is `final`. Changing our minds from no type hint to a string type hint will not be a BC-break, but the reverse will be. So I say let's be strict for now, and if it turns out we should be lax later, then we will be able to thanks to `final`.

leroy0211: Sure! Lets do that.

src/Twig/Extension/SeoExtension.php
leroy0211: I've also updated all the html rendering to a block-type way (just like symfony forms) in a separate commit. Should I make a different PR for this? Or should I add it in here.

leroy0211:
> Looks good, preventing merge until the todo list is done though.

Yep, I was waiting to alter the docs until the changes have been accepted.

src/Seo/AttributeBag.php
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\SeoBundle\Seo;
+
+final class AttributeBag implements \IteratorAggregate
+{
+    private $attributes = [];
```
core23: You could add a PHPDoc here:
```suggestion
   /**
     * @param array<string|string>
     */
    private $attributes = [];
```

greg0ire:
@leroy0211 what's the status of your PRs?

core23:
Ping @leroy0211

leroy0211:
Sorry, hadn't had the time to update my PR's in a while now. If anyone would like to collaborate?

PR #330, opened 2019-02-12
Cast metadata to string

pierredup:
## Subject

Strict types were added in the latest version (#313), which causes metadata values to not be implicitly converted to strings anymore.
Most notably what I picked up is the [currency in the ProductBundle](https://github.com/sonata-project/ecommerce/blob/2.x/src/ProductBundle/Seo/Services/Facebook.php#L96) which is an object (containing a `__toString` method) and the [image dimensions for a product](https://github.com/sonata-project/ecommerce/blob/2.x/src/ProductBundle/Seo/Services/Facebook.php#L113-L114) which are integers.

I am targeting this branch, because it is a bug fix introduced in the latest stable version.

## Changelog

```markdown
### Fixed
- cast metadata to string during render
```

### DIFF
--- a/src/Twig/Extension/SeoExtension.php
+++ b/src/Twig/Extension/SeoExtension.php
@@ -262,6 +262,6 @@ class SeoExtension extends \Twig_Extension
      */
     private function normalize($string)
     {
-        return htmlentities(strip_tags($string), ENT_COMPAT, $this->encoding);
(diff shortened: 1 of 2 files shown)


pierredup:
Done in #331

greg0ire:
> Done in #331

Nice!

PR #331, opened 2019-02-13
Deprecate passing non-string meta content

pierredup:
## Subject

This is a follow-up to #330, to deprecate passing non-string values for meta content.

## Changelog

```markdown
### Deprecated
- passing a non-string value when adding meta content
```

### DIFF
--- a/src/Seo/SeoPage.php
+++ b/src/Seo/SeoPage.php
@@ -119,8 +119,16 @@ class SeoPage implements SeoPageInterface
     /**
      * {@inheritdoc}
      */
-    public function addMeta($type, $name, $content, array $extras = [])
+    public function addMeta($type, $name, /* string */ $content, array $extras = [])
     {
+        if (!\is_string($content)) {
+            @trigger_error(sprintf(
+                'Passing meta content of type %s in %s is deprecated since version 2.x and will be unsupported in version 3. Please cast the value to a string first.',
+                 \gettype($content),
+                  __METHOD__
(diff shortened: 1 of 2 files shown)


src/Seo/SeoPage.php
```diff
@@ -119,8 +119,12 @@ public function getMetas()
     /**
      * {@inheritdoc}
      */
-    public function addMeta($type, $name, $content, array $extras = [])
+    public function addMeta($type, $name, /* string */ $content, array $extras = [])
     {
+        if (!is_string($content)) {
+            @trigger_error(sprintf('Passing meta content of type %s in %s is deprecated since version 2.7.1 and will be unsupported in version 3. Please cast the value to a string first.', gettype($content), __METHOD__), E_USER_DEPRECATED);
```
```suggestion
            @trigger_error(sprintf(
                'Passing meta content of type %s in %s is deprecated since version 2.x and will be unsupported in version 3. Please cast the value to a string first.',
                 gettype($content),
                  __METHOD__
              ), E_USER_DEPRECATED);
```

greg0ire:
I think you will have to change this to fix the build: https://github.com/sonata-project/SonataSeoBundle/blob/b73bf69675c1657d4217b416247ef667a6ec6c2a/tests/Seo/SeoPageTest.php#L157

OskarStark:
@Soullivaneuh is FlintCI working properly? 😐

greg0ire:
Thanks @pierredup !

PR #336, opened 2019-05-17
Twig deprecation notice

emulienfou:
## Subject

Fixing Twig deprecation notice in Symfony 4

## Changelog
```markdown
### Fixed
* deprecation notice about using namespaced classes from `\Twig\`
```

### DIFF
--- a/src/Twig/Extension/SeoExtension.php
+++ b/src/Twig/Extension/SeoExtension.php
@@ -14,8 +14,10 @@ declare(strict_types=1);
 namespace Sonata\SeoBundle\Twig\Extension;
 
 use Sonata\SeoBundle\Seo\SeoPageInterface;
+use Twig\Extension\AbstractExtension;
+use Twig\TwigFunction;
 
-class SeoExtension extends \Twig_Extension
+class SeoExtension extends AbstractExtension
 {
     /**
      * @var SeoPageInterface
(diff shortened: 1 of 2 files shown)


src/Twig/Extension/SeoExtension.php
```diff
@@ -14,8 +14,10 @@
 namespace Sonata\SeoBundle\Twig\Extension;
 
 use Sonata\SeoBundle\Seo\SeoPageInterface;
+use Twig\Extension\AbstractExtension;
+use Twig\TwigFunction;
 
-class SeoExtension extends \Twig_Extension
+class SeoExtension extends AbstractExtension
```
kunicmarko20: can we do this?

emulienfou: Yes we can, just tested on a real app.
For information Twig namespaces are available since Twig 1.34+ and 2.4+

kunicmarko20: I was thinking about BC breaks

emulienfou: IMHO, since I update minimum version of twig from **2.0** to **2.4** this should not be an issue, no?

greg0ire: I think this is not a BC break because one class is an alias for the other one, right?

greg0ire:
Please associate your email address with your Github account, or change the email in your commits to an address already associated with it.

greg0ire:
Please kindly squash your commits together. If you don't, we'll try to remember to do it for you but it's best if you save us this trouble.

How to do that?

1. `git rebase -i origin/3.x`, assuming `origin` is a git remote that points to this repository, and not your fork. If you're not sure what your remotes are, run `git remote -vvv`, there should be your fork and the holy/reference/base/origin/whatever-you-call-it repository.
2. A window will show up with many lines, replace `pick` with `fixup` on every line but the first one
3. Close your editor, git should do its magic, and you should end up with one commit
4. Use `git push --force` to overwrite what you already push. Don't forget the `--force` option otherwise git will try to merge both things together.

greg0ire:
Blocked by #337

greg0ire:
Closing and reopening is not enough apparently, please rebase.

src/Twig/Extension/SeoExtension.php
emulienfou: Exactly, there is still a class alias in the namespaced version: `class_alias('Twig\Extension\AbstractExtension', 'Twig_Extension');`

emulienfou:
> Wound you mind check the other bundles for the same changes if you are on it? This would help us a lot

@OskarStark already did the same PR on this bundle: https://github.com/liip/LiipThemeBundle/pull/196

PR #361, opened 2019-12-06
Deprecate Guzzle with psr http-client

core23:
I am targetting this branch, because this adds a new implementation without removing the existing one. This is a rework of #161, but targets the psr http-client.

## Changelog

``` markdown
### Added
- Added support for `psr/http-client` in `TwitterEmbedTweetBlockService`
```
## Subject

Added support for an abstract http client without removing the existing old guzzle implementation.

### DIFF
--- a/UPGRADE-2.x.md
+++ b/UPGRADE-2.x.md
@@ -1,6 +1,13 @@
 UPGRADE 2.x
 ===========
 
+## TwitterEmbedTweetBlockService uses psr/http-client
+
+The `Guzzle` dependency was removed and replaced with abstract `http-client` interface, so you can choose your preferred
+client implementation. If you have extended the `TwitterEmbedTweetBlockService` class, you need to adjust the
+constructor signature.
+
+
 UPGRADE FROM 2.2 to 2.3
 =======================
 
@@ -23,6 +30,6 @@ UPGRADE FROM 2.0 to 2.1
 
 ### Tests
 
-All files under the ``Tests`` directory are now correctly handled as internal test classes. 
-You can't extend them anymore, because they are only loaded when running internal tests. 
+All files under the ``Tests`` directory are now correctly handled as internal test classes.
+You can't extend them anymore, because they are only loaded when running internal tests.
 More information can be found in the [composer docs](https://getcomposer.org/doc/04-schema.md#autoload-dev).
--- a/composer.json
+++ b/composer.json
@@ -21,6 +21,8 @@
     ],
     "require": {
         "php": "^7.1",
+        "psr/http-client": "^1.0",
+        "psr/http-message": "^1.0",
         "sonata-project/exporter": "^1.11 || ^2.0",
         "symfony/config": "^3.4 || ^4.2 || ^5.0",
         "symfony/dependency-injection": "^3.4 || ^4.2 || ^5.0",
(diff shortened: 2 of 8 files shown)


docs/reference/social_blocks.rst
```diff
@@ -49,5 +49,12 @@ The ``sonata.seo.block.twitter.embed`` allows you to embed a tweet by giving its
 or embed content. Please refer to `Embedded Tweets doc <https://dev.twitter.com/docs/embedded-tweets>`_
 and `OEmbed API doc <https://dev.twitter.com/docs/api/1/get/statuses/oembed>`_ for a full understanding.
 
-The block service allows you to ask the API if you've given an ID or a URL (you'll need to install the
-Guzzle library for this: ``composer require guzzle/guzzle``).
+The block service allows you to ask the API if you've given an ID or a URL .
+
+You'll need to install a `psr/http-client` (e.g. `symfony/http-client`) and a `psr/http-message` (e.g `nyholm/psr7`)
```
greg0ire:
```suggestion
You'll need to install a ``psr/http-client`` (e.g. ``symfony/http-client``) and a ``psr/http-message`` (e.g ``nyholm/psr7``)
```

docs/reference/social_blocks.rst
```diff
@@ -49,5 +49,12 @@ The ``sonata.seo.block.twitter.embed`` allows you to embed a tweet by giving its
...
+to request the Twitter API:
+
+.. code-block:: bash
+
+    composer require symfony/http-client
```
greg0ire: It's safer to require all packages at once, in case the latest version of the second one is not compatible yet with the latest version of the first one.

```suggestion
    composer require symfony/http-client nyholm/psr7
```

UPGRADE-2.x.md
```diff
@@ -1,6 +1,13 @@
 UPGRADE 2.x
 ===========
 
+## TwitterEmbedTweetBlockService uses psr/http-client
+
+The `Guzzle` dependency was removed and replaced with abstract `http-client` interface, so you can choose your preferred
+client implementation. If you have extended the `TwitterEmbedTweetBlockService` class, you need to adjust the
+constructor signature.
```
greg0ire: So… this is a BC-break?

core23: No, both ways are possible

greg0ire: Does not look like it: https://3v4l.org/iUNnq

core23: How is this related to the docs? 😕

src/Block/Social/TwitterEmbedTweetBlockService.php
```diff
@@ -41,36 +45,32 @@ class TwitterEmbedTweetBlockService extends BaseTwitterButtonBlockService
...
-                $tweet = $apiTweet['html'];
-            } catch (CurlException $e) {
-                // log error
-            }
-        }
+        $this->httpClient = $httpClient;
+        $this->messageFactory = $messageFactory;
+    }
 
+    public function execute(BlockContextInterface $blockContext, Response $response = null): Response
```
greg0ire: @core23 I though the docs were talking about this change which adds a return type declaration that was not there before… this `: Reponse` is a BC break, isn't it?

UPGRADE-2.x.md
greg0ire: Oh right, didn't see it was about the ctor, my bad.

composer.json
```diff
@@ -35,19 +37,21 @@
...
-        "guzzle/guzzle": "^3.8",
+        "guzzlehttp/guzzle": "^6.0",
+        "nyholm/psr7": "^1.2",
         "sonata-project/admin-bundle": "^3.31",
         "sonata-project/block-bundle": "^3.11",
         "sonata-project/core-bundle": "^3.9",
         "symfony/console": "^3.4 || ^4.2 || ^5.0",
         "symfony/filesystem": "^3.4 || ^4.2 || ^5.0",
         "symfony/finder": "^3.4 || ^4.2 || ^5.0",
+        "symfony/http-client": "^4.3 || ^4.2 || ^5.0",
```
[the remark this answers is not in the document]
OskarStark: Why not require 4.4 here instead of 4.3 ?

greg0ire:
Thanks @core23 !

PR #399, opened 2020-06-06
Bump block-bundle dependency

core23:
## Subject

I am targeting this branch, because the parent class is changed and it cannot be done without breaking BC.

## Changelog

```markdown
### Changed
- Bump block-bundle dependency
- Changed parent class of all block / breadcrumb services
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,21 +36,11 @@ matrix:
     - php: '7.3'
       env: SYMFONY='dev-master as 4.4.x-dev'
     - php: '7.3'
-      env: SONATA_BLOCK=3.*
-    - php: '7.3'
-      env: SONATA_BLOCK='dev-master as 3.x-dev'
-    - php: '7.3'
-      env: SONATA_ADMIN=3.*
-    - php: '7.3'
-      env: SONATA_ADMIN='dev-master as 3.x-dev'
-    - php: '7.3'
       env: SYMFONY_DEPRECATIONS_HELPER=0
   allow_failures:
     - php: nightly
     - env: SYMFONY_DEPRECATIONS_HELPER=0
     - env: SYMFONY='dev-master as 4.4.x-dev'
-    - env: SONATA_BLOCK='dev-master as 3.x-dev'
-    - env: SONATA_ADMIN='dev-master as 3.x-dev'
 
 before_install:
   - git remote add upstream ${UPSTREAM_URL} && git fetch --all
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,6 @@ branches:
 language: php
 
 php:
-  - '7.2'
   - '7.3'
   - '7.4'
   - nightly
@@ -30,7 +29,7 @@ env:
 matrix:
   fast_finish: true
   include:
-    - php: '7.2'
+    - php: '7.3'
       env: COMPOSER_FLAGS="--prefer-lowest"
     - php: '7.3'
       env: SYMFONY=4.4.*
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
         }
     ],
     "require": {
-        "php": "^7.1",
+        "php": "^7.3",
         "psr/http-client": "^1.0",
         "psr/http-message": "^1.0",
         "sonata-project/block-bundle": "^4.0",
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,10 @@
         "php": "^7.1",
         "psr/http-client": "^1.0",
         "psr/http-message": "^1.0",
+        "sonata-project/block-bundle": "^4.0",
+        "sonata-project/doctrine-extensions": "^1.6",
         "sonata-project/exporter": "^1.11 || ^2.0",
+        "sonata-project/form-extensions": "^1.0",
         "symfony/config": "^4.4 || ^5.0",
         "symfony/dependency-injection": "^4.4 || ^5.0",
         "symfony/form": "^4.4 || ^5.0",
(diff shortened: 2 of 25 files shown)


composer.json
```diff
@@ -23,7 +23,9 @@
         "php": "^7.1",
         "psr/http-client": "^1.0",
         "psr/http-message": "^1.0",
+        "sonata-project/block-bundle": "^4.0",
```
jordisala1991, line 26: Should we move this to require dev, like before?

core23: I'm not a big fan of optional dependencies, because they often cause trouble. E.g. the 3.x release of this bundle is incompatible with the block 4.0 release, but you can install both packages inside a project.

The latest major block bundle does not bring too much 3rd party libs that are not already required by this bundle.

wbloszyk:
Upgrade BlockBundle from 3 to 4 contains many changes like other API, close API, return Types hints. We should consider(i think we have to) add support for both version. Otherwise we will be not able to upgrade bundles one by one.

core23:
> We should consider(i think we have to) add support for both version

This is not possible without, because we change the base class of too many services. The old base classes are deprecated and do not exist in block bundle 4.x anymore.

wbloszyk:
I think more about including both Block version for each bundle(Block3 and Block4 directory) and create aliases depends on register BlockBundle version(Block directory).

core23:
> [...] and create aliases depends on register BlockBundle version(Block directory).

Feel free to do this, but this PR could not be solved without using some dirty hacks, because we change the base class of the `BaseBreadcrumbMenuBlockService`. I would not use too much class proxies, because they are hard to maintain and will cause some major issues later.

core23:
Ping @sonata-project/contributors

core23:
Please review @sonata-project/contributors

.travis.yml
```diff
@@ -10,7 +10,6 @@ branches:
 language: php
 
 php:
-  - '7.2'
```
franmomu, line 13: I guess https://github.com/sonata-project/dev-kit/blob/master/config/projects.yml#L412 should be updated

src/Block/Social/FacebookLikeButtonBlockService.php
```diff
@@ -104,9 +112,13 @@ public function buildEditForm(FormMapper $formMapper, BlockInterface $block): vo
...
 
-    public function getBlockMetadata($code = null)
+    public function validate(ErrorElement $errorElement, BlockInterface $block): void
+    {
+    }
+
+    public function getMetadata(): MetadataInterface
     {
-        return new Metadata($this->getName(), (null !== $code ? $code : $this->getName()), false, 'SonataSeoBundle', [
+        return new Metadata('onata.seo.block.facebook.like_button', null, null, 'SonataSeoBundle', [
```
franmomu, line 121:
```suggestion
        return new Metadata('sonata.seo.block.facebook.like_button', null, null, 'SonataSeoBundle', [
```

src/Block/Social/FacebookSendButtonBlockService.php
```diff
@@ -42,7 +45,12 @@ public function configureSettings(OptionsResolver $resolver): void
         ]);
     }
 
-    public function buildEditForm(FormMapper $formMapper, BlockInterface $block): void
+    public function configureCreateForm(\Sonata\BlockBundle\Form\Mapper\FormMapper $formMapper, BlockInterface $block): void
```
franmomu, line 48:
```suggestion
    public function configureCreateForm(FormMapper $formMapper, BlockInterface $block): void
```

src/Block/Social/FacebookSendButtonBlockService.php
```diff
@@ -68,9 +76,13 @@ public function buildEditForm(FormMapper $formMapper, BlockInterface $block): vo
...
 
-    public function getBlockMetadata($code = null)
+    public function validate(ErrorElement $errorElement, BlockInterface $block): void
+    {
+    }
+
+    public function getMetadata(): MetadataInterface
     {
-        return new Metadata($this->getName(), (null !== $code ? $code : $this->getName()), false, 'SonataSeoBundle', [
+        return new Metadata('onata.seo.block.facebook.send_button', null, null, 'SonataSeoBundle', [
```
franmomu, line 85:
```suggestion
        return new Metadata('sonata.seo.block.facebook.send_button', null, null, 'SonataSeoBundle', [
```

src/Block/Social/FacebookShareButtonBlockService.php
```diff
@@ -53,7 +55,12 @@ public function configureSettings(OptionsResolver $resolver): void
         ]);
     }
 
-    public function buildEditForm(FormMapper $formMapper, BlockInterface $block): void
+    public function configureCreateForm(\Sonata\BlockBundle\Form\Mapper\FormMapper $formMapper, BlockInterface $block): void
```
franmomu, line 58:
```suggestion
    public function configureCreateForm(FormMapper $formMapper, BlockInterface $block): void
```

src/Block/Social/TwitterShareButtonBlockService.php
```diff
@@ -48,7 +51,12 @@ public function configureSettings(OptionsResolver $resolver): void
         ]);
     }
 
-    public function buildEditForm(FormMapper $formMapper, BlockInterface $block): void
+    public function configureCreateForm(\Sonata\BlockBundle\Form\Mapper\FormMapper $formMapper, BlockInterface $block): void
```
franmomu, line 54:
```suggestion
    public function configureCreateForm(FormMapper $formMapper, BlockInterface $block): void
```

.travis.yml
core23: https://github.com/sonata-project/dev-kit/pull/775

PR #450, opened 2020-10-12
Bump block dependency

core23:
## Subject

## Changelog

```markdown
### Changed
- Bump block dependency
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
         }
     ],
     "require": {
-        "php": "^7.2",
+        "php": "^7.3",
         "psr/http-client": "^1.0",
         "psr/http-message": "^1.0",
         "sonata-project/exporter": "^2.0",
--- a/composer.json
+++ b/composer.json
@@ -34,14 +34,14 @@
         "twig/twig": "^2.9 || ^3.0"
     },
     "conflict": {
-        "sonata-project/block-bundle": "<3.11",
+        "sonata-project/block-bundle": "<4.4",
         "sonata-project/core-bundle": "<3.20"
     },
     "require-dev": {
         "guzzlehttp/guzzle": "^6.0",
         "nyholm/psr7": "^1.2",
         "sonata-project/admin-bundle": "^3.31",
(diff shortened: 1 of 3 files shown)


src/Block/Breadcrumb/BaseBreadcrumbMenuBlockService.php
```diff
@@ -74,7 +74,7 @@ public function getName()
     /**
      * {@inheritdoc}
      */
-    public function configureSettings(OptionsResolver $resolver)
+    public function configureSettings(OptionsResolver $resolver): void
```
core23, line 77: Technically this is a BC break, but if you are using sonata block bundle 4, you MUST use this signature.

composer.json
```diff
@@ -34,14 +34,14 @@
...
-        "sonata-project/block-bundle": "<3.11",
+        "sonata-project/block-bundle": "<4.4",
         "sonata-project/core-bundle": "<3.20"
     },
     "require-dev": {
         "guzzlehttp/guzzle": "^6.0",
         "nyholm/psr7": "^1.2",
         "sonata-project/admin-bundle": "^3.31",
-        "sonata-project/block-bundle": "^3.17",
+        "sonata-project/block-bundle": "^4.4",
```
jordisala1991, line 44: The other bundles are allowing 3 or 4

core23: We can't do this, because we would risk a BC break in `BaseBreadcrumbMenuBlockService::configureSettings`

VincentLanglet (dismissed):

composer.json
VincentLanglet: Without BlockBundle 3, the test can't run on php 7.2.

Is there a reason for dropping 7.2 support on BlockBundle 4 ?

jordisala1991: All new major releases are dropping 7.2

IMO we could study what is the impact of the BC break if supporting sonata block 3, and if it is not too big, just accept it.

core23:
Doing this on the stable branch will result in many BC breaks :(

PR #511, opened 2021-02-20
Update phpdoc 

core23:
## Subject

## Changelog

```markdown
### Added
- Added generic information to `SeoPageInterface`

### Changed
- Soft move methods to `SeoPageInterface` using @method
```

### DIFF
--- a/src/Seo/SeoPage.php
+++ b/src/Seo/SeoPage.php
@@ -24,12 +24,12 @@ class SeoPage implements SeoPageInterface
     protected $title;
 
     /**
-     * @var array
+     * @var array<string, array<string, mixed>>
      */
     protected $metas;
 
     /**
-     * @var array
+     * @var array<string, string>
      */
     protected $htmlAttributes;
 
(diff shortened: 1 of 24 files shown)


VincentLanglet (dismissed):

src/Seo/SeoPage.php
```diff
@@ -184,222 +154,139 @@ public function setMetas(array $metadatas)
...
-    /**
-     * @return array
-     */
     public function getOEmbedLinks()
     {
         return $this->oembedLinks;
     }
 
     /**
      * @param mixed $meta
```
franmomu, line 287:
```suggestion
     * @param string|array $meta
```
based on the code, looks like it could be defined this way.

franmomu (approved):

src/Seo/SeoPage.php
core23: Sure? What about integer or boolean values?

franmomu: I mean, the code is like this:
```php
    /**
     * @param mixed $meta
     */
    private function normalize($meta): array
    {
        if (\is_string($meta)) {
            return [$meta, []];
        }

        return $meta;
    }
```
with other scalar value it will throw a `TypeError`.

franmomu (approved):

franmomu:
thanks @core23!

Issue #540, opened 2021-05-23
The service "sonata.seo.block.email.share_button" has a dependency on a non-existent service "sonata.templating"


patrickmaynard:
#### Sonata packages

<details><summary>show</summary>
<p>

```
$ composer show --latest 'sonata-project/*'
sonata-project/admin-bundle              3.100.1 3.100.1 The missing Symfony ...
sonata-project/block-bundle              4.5.3   4.5.3   Symfony SonataBlockB...
sonata-project/cache                     2.1.1   2.1.1   Cache library
sonata-project/cache-bundle              3.3.0   3.3.0   This bundle provides...
sonata-project/datagrid-bundle           3.3.0   3.3.0   Symfony SonataDatagr...
sonata-project/doctrine-extensions       1.12.0  1.12.0  Doctrine2 behavioral...
sonata-project/doctrine-orm-admin-bundle 3.34.1  3.34.1  Integrate Doctrine O...
sonata-project/exporter                  2.6.2   2.6.2   Lightweight Exporter...
sonata-project/form-extensions           1.9.0   1.9.0   Symfony form extensions
sonata-project/seo-bundle                2.13.0  2.13.0  Symfony SonataSeoBundle
sonata-project/twig-extensions           1.6.0   1.6.0   Sonata twig extensions
sonata-project/user-bundle               4.11.1  4.11.1  Symfony SonataUserBu...
```

</p>
</details>

#### Symfony packages

<details><summary>show</summary>
<p>

```
$ composer show --latest 'symfony/*'
symfony/acl-bundle                 v2.1.0  v2.1.0  Symfony AclBundle
symfony/asset                      v5.2.7  v5.2.7  Manages URL generation and versioning o...
symfony/cache                      v5.2.9  v5.2.9  Provides an extended PSR-6, PSR-16 (and...
symfony/cache-contracts            v2.4.0  v2.4.0  Generic abstractions related to caching
symfony/config                     v4.4.23 v5.2.8  Helps you find, load, combine, autofill...
symfony/console                    v4.4.24 v5.2.8  Eases the creation of beautiful and tes...
symfony/debug                      v4.4.22 v4.4.22 Provides tools to ease debugging PHP code
symfony/dependency-injection       v4.4.24 v5.2.9  Allows you to standardize and centraliz...
symfony/deprecation-contracts      v2.4.0  v2.4.0  A generic function and convention to tr...
[...]

PR #541, opened 2021-05-25
Fix to issue #540

patrickmaynard:
## Subject

Fix block bundle requirement

I am targeting this branch, because it is a clarification of an already-existing dependency, rather than a breaking change. (Currently, if a block bundle outside of the range I'm specifying is installed, the SEO bundle malfunctions. So this is a clarification, rather than something that significantly changes existing behavior.)

Closes ##540.

### Changed

-- Commit one: Add an acceptable version range for the block bundle to composer.json
-- Commit two: Fix missing comma in composer.json

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
         "symfony/http-foundation": "^4.4 || ^5.0",
         "symfony/http-kernel": "^4.4 || ^5.0",
         "symfony/options-resolver": "^4.4 || ^5.0",
-        "twig/twig": "^2.9 || ^3.0"
+        "twig/twig": "^2.9 || ^3.0", 
         "sonata-project/block-bundle": ">=3.0 <4.0"
     },
     "conflict": {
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -31,7 +31,8 @@
         "symfony/http-foundation": "^4.4 || ^5.0",
         "symfony/http-kernel": "^4.4 || ^5.0",
         "symfony/options-resolver": "^4.4 || ^5.0",
-        "twig/twig": "^2.9 || ^3.0"
+        "twig/twig": "^2.9 || ^3.0", 
+        "sonata-project/block-bundle": ">=3.0 <4.0"
```
VincentLanglet, line 35:
```suggestion
        "sonata-project/block-bundle": "^3.11"
```
And you can remove it from the conflict section then.

And also, the dependencies have to be alphabetically ordered

composer.json
```diff
@@ -31,7 +31,8 @@
...
     },
     "conflict": {
         "sonata-project/block-bundle": "<3.11",
```
core23, line 38: You can remove this conflict and require-dev entry when adding this as a required dependency

core23 (changes requested):

patrickmaynard:
Thanks for the comments. Since there are a couple problems with pipelines running on this version, I'm closing this pull request and opening another. 

https://github.com/sonata-project/SonataSeoBundle/pull/542

I have integrated your comments on that request.

PR #354, opened 2019-11-29
Add support for Symfony 5 and Twig 3

kuraobi:
## Add support for Symfony 5 and Twig 3

I guess this one is self explanatory...

I am targeting this branch, because it only adds support for newer versions of Twig and Symfony.

Closes #353

## Changelog

```markdown
### Added
- Support for Twig 3 and Symfony 5
```

PR #548, opened 2021-08-05
Bump dependencies

VincentLanglet:


jordisala1991 (approved):

jordisala1991:
Maybe this one needs a changelog?

jordisala1991:
BTW there is no build for sf 5.3

composer.json
```diff
@@ -25,13 +25,13 @@
...
-        "symfony/dependency-injection": "^4.4 || ^5.0",
-        "symfony/form": "^4.4 || ^5.0",
-        "symfony/framework-bundle": "^4.4 || ^5.0",
-        "symfony/http-foundation": "^4.4 || ^5.0",
-        "symfony/http-kernel": "^4.4 || ^5.0",
-        "symfony/options-resolver": "^4.4 || ^5.0",
+        "symfony/config": "^4.4 || ^5.3",
+        "symfony/dependency-injection": "^4.4 || ^5.3",
+        "symfony/form": "^4.4 || ^5.3",
+        "symfony/framework-bundle": "^4.4 || ^5.3",
```
jordisala1991, line 31: this dependency is not getting installed in any build.

VincentLanglet:
> BTW there is no build for sf 5.3

We should add one

composer.json
VincentLanglet: It was added here: https://github.com/sonata-project/SonataSeoBundle/pull/354

But seems like we're not fully supporting Symfony 5.

VincentLanglet:
I think we cannot add a SF5 build because it use SonataAdmin v3 !

jordisala1991:
> I think we cannot add a SF5 build because it use SonataAdmin v3 in require-dev !

IMO we should revert what cannot be installed with SonataAdmin v3, and leave the full support for next major.

VincentLanglet:
> > I think we cannot add a SF5 build because it use SonataAdmin v3 in require-dev !
> 
> IMO we should revert what cannot be installed with SonataAdmin v3, and leave the full support for next major.

Not really because you're supposed to be able to use this without SonataAdmin I think.
It's just that we're not testing this bundle without SonataAdmin.

jordisala1991:
Yep, I agree with that, but that's the philosphy we took with all sonata bundles. If the pipelines does not pass with symfony 5 installed, you can't add support.

VincentLanglet:
> Yep, I agree with that, but that's the philosphy we took with all sonata bundles. If the pipelines does not pass with symfony 5 installed, you can't add support.

I'm not sure about this.
We refused it for `require` but not `require-dev`. For instance `php8` support were added for some repository before all the require-dev sections were supporting it.

jordisala1991:
> > Yep, I agree with that, but that's the philosphy we took with all sonata bundles. If the pipelines does not pass with symfony 5 installed, you can't add support.
> 
> I'm not sure about this.
> We refused it for `require` but not `require-dev`. For instance `php8` support were added for some repository before all the require-dev sections were supporting it.

For php8 we used the --ignore-platform-reqs to start testing it on github actions. But we didn't add it to composer.json until it was installable without that flag, you can check the remaining repositories without support for php 8. All of them are working with php 8.0 (ignoring the requirements).

VincentLanglet:
> For php8 we used the --ignore-platform-reqs to start testing it on github actions. But we didn't add it to composer.json until it was installable without that flag, you can check the remaining repositories without support for php 8.

Here it would be the same to say that if you remove `admin-bundle`, non-admin-related tests works.
Right thing to do would be to skip admin-related test if admin-bundle is not installed and to add a CI to run tests without admin-bundle. Because without a proof that it doesn't work, people with Symfony5 and this bundle won't be able to update if we remove the compatibility in the composer.json.

jordisala1991:
> > For php8 we used the --ignore-platform-reqs to start testing it on github actions. But we didn't add it to composer.json until it was installable without that flag, you can check the remaining repositories without support for php 8.
> 
> Here it would be the same to say that if you remove `admin-bundle`, non-admin-related tests works.
> Right thing to do would be to skip admin-related test if admin-bundle is not installed and to add a CI to run tests without admin-bundle. Because without a proof that it doesn't work, people with Symfony5 and this bundle won't be able to update if we remove the compatibility in the composer.json.

Phpstan wouldn't let you remove admin-bundle from composer.json, not without disable part of the checks, which is not a good idea.

To do what you say we would need to do an extra build removing all "trully optional dependencies", to check the bundle is functional and installs symfony 5.

Tbh I like the idea, but what I said about not accepting PRs that do not bring full support was applied here: https://github.com/sonata-project/SonataMediaBundle/blob/3.x/composer.json (probably on other bundles too)

VincentLanglet:
> Phpstan wouldn't let you remove admin-bundle from composer.json, not without disable part of the checks, which is not a good idea.

The more interesting would be the phpunit tests.

> Tbh I like the idea, but what I said about not accepting PRs that do not bring full support was applied here: https://github.com/sonata-project/SonataMediaBundle/blob/3.x/composer.json (probably on other bundles too)

In the mediaBundle it wasn't added still.
Here, it was added, against your (our ?) policy. Removing it now could be kinda considered as a regression.
People will be stuck will old versions.

Speaking about SonataMediaBundle, I remembered that recently someone ask for support of Symfony 5 on slack. And when I answered that we first need to add support for SonataAdmin 4 ; he answered "I'm not using SonataAdmin", just "SonataMedia" and it works fine. I didn't like the fact to block him because we want "All or nothing"...

Here, if you look at the SonataAdminBundle usage, 
https://github.com/sonata-project/SonataSeoBundle/search?p=1&q=AdminBundle
we're using only one thing `use Sonata\AdminBundle\Form\FormMapper;`.

Best thing would be to add support for BlockBundle 4, in order to be able to remove the SonataAdmin dependency.

jordisala1991:
> > Phpstan wouldn't let you remove admin-bundle from composer.json, not without disable part of the checks, which is not a good idea.
> 
> The more interesting would be the phpunit tests.
> 
> > Tbh I like the idea, but what I said about not accepting PRs that do not bring full support was applied here: https://github.com/sonata-project/SonataMediaBundle/blob/3.x/composer.json (probably on other bundles too)
> 
> In the mediaBundle it wasn't added still.
> Here, it was added, against your (our ?) policy. Removing it now could be kinda considered as a regression.
> People will be stuck will old versions.
> 
> Speaking about SonataMediaBundle, I remembered that recently someone ask for support of Symfony 5 on slack. And when I answered that we first need to add support for SonataAdmin 4 ; he answered "I'm not using SonataAdmin", just "SonataMedia" and it works fine. I didn't like the fact to block him because we want "All or nothing"...
> 
> Here, if you look at the SonataAdminBundle usage,
> https://github.com/sonata-project/SonataSeoBundle/search?p=1&q=AdminBundle
> we're using only one thing `use Sonata\AdminBundle\Form\FormMapper;`.
> 
> Best thing would be to add support for BlockBundle 4, in order to be able to remove the SonataAdmin dependency.

Well, even if I am showing that we do not even know if the bundle works with symfony 5, because we have no build testing it, it is not my policy, we have been doing this way for all symfony 5 support so far (this could be the only exception with a very old pr giving support for sf 5).

In fact I do agree on not blocking sf support for require-dev dependencies, but we must have a way to test it, otherwise we are blind.

VincentLanglet:
> In fact I do agree on not blocking sf support for require-dev dependencies, but we must have a way to test it, otherwise we are blind.

Yes.

Short way, we can add a github actions here.

Generic way, we should find a way to remove a bundle on a build during tests. Same way we have a `Test / PHP 8.0 + highest + sonata-project/block-bundle:"3.*" (pull_request)` build to specify a version. I would be possible by improving
https://github.com/sonata-project/dev-kit/blob/master/templates/project/.github/workflows/test.yaml.twig#L98-L100

Long way, adding support for block bundle 4 since adminBundle won't be require.

PR #579, opened 2021-09-14
Increase phpstan level

core23:
## Subject

Fix minor stan issues.


phpstan.neon.dist
```diff
@@ -1,5 +1,5 @@
 parameters:
```
VincentLanglet, line 1:
```
includes:
    - phpstan-baseline.neon

parameters:
    level: 5
    bootstrapFiles:
        - vendor/bin/.phpunit/phpunit/vendor/autoload.php
    paths:
        - src
        - tests
    treatPhpDocTypesAsCertain: false
```

?

VincentLanglet (commented):
You need the psalm-symfony plugin and the phpstan-symfony plugin IMHO

core23:
Depends on #580 #585

phpstan-baseline.neon
```diff
@@ -1,3 +1,7 @@
 parameters:
 	ignoreErrors:
+		-
+			message: "#^Parameter \\#1 \\$metas of method Sonata\\\\SeoBundle\\\\Seo\\\\SeoPage\\:\\:setMetas\\(\\) expects array\\<string, array\\<string, mixed\\>\\>, array\\<string, string\\> given\\.$#"
```
VincentLanglet: I would say we can fix this

core23: No idea how to fix this. This is a test issue where we ignore the type hints.

VincentLanglet: I see, this is the call 
```
public function testInvalidMetas()
    {
        $this->expectException(\RuntimeException::class);

        $page = new SeoPage();
        $page->setMetas([
            'type' => 'not an array',
        ]);
    }
```

Indeed we cannot fix it, but on other project we ignored the error directly in the file with a psalm-suppress and phpstan-ignore https://github.com/sonata-project/SonataAdminBundle/blob/8a4e7f5210353c25b80614f2dec20bf4f7217ffe/tests/FieldDescription/TypeGuesserChainTest.php#L28

Kinda using the baseline as a TODO and this error will never be fixed.

core23:
phpstan is failing in the ci, because the docker action is using an old version of the phpstan-symfony extension.

cc @OskarStark

VincentLanglet:
> phpstan is failing in the ci, because the docker action is using an old version of the phpstan-symfony extension.
> 
> cc @OskarStark

You should remove this ci build https://github.com/sonata-project/SonataSeoBundle/blob/3.x/.github/workflows/qa.yml
There is already https://github.com/sonata-project/SonataSeoBundle/blob/3.x/.github/workflows/qa.yaml

core23:
> > phpstan is failing in the ci, because the docker action is using an old version of the phpstan-symfony extension.
> > cc @OskarStark
> 
> You should remove this ci build https://github.com/sonata-project/SonataSeoBundle/blob/3.x/.github/workflows/qa.yml
> There is already https://github.com/sonata-project/SonataSeoBundle/blob/3.x/.github/workflows/qa.yaml

Good catch! 👍

VincentLanglet (dismissed):

src/DependencyInjection/Compiler/ServiceCompilerPass.php
```diff
@@ -33,9 +33,8 @@ public function process(ContainerBuilder $container): void
         $definition->addMethodCall('setHtmlAttributes', [$config['head']]);
         $definition->addMethodCall('setSeparator', [$config['separator']]);
 
-        if ($alias = $container->setAlias('sonata.seo.page', $config['default'])) {
-            $alias->setPublic(true);
-        }
+        $alias = $container->setAlias('sonata.seo.page', $config['default']);
```
jordisala1991: no need this otv, you can directly call setPublic

src/DependencyInjection/SonataSeoExtension.php
```diff
@@ -103,6 +107,10 @@ private function configureSitemap(array $config, ContainerBuilder $container): v
 
     /**
      * Fix the sitemap configuration.
+     *
+     * @param array<mixed> $config
```
jordisala1991: mixed[]

tests/DependencyInjection/ConfigurationTest.php
```diff
@@ -100,6 +103,11 @@ private function getDefaultConfiguration(): array
         ];
     }
 
+    /**
+     * @param array<array-key, mixed> $configs
+     *
+     * @return array<array-key, mixed>
```
jordisala1991: mixed[], but I would recommend you to look at how it is done on AdminBundle and copy the same here

tests/Seo/SeoPageTest.php
```diff
@@ -55,11 +55,15 @@ public function testRemoveMeta(): void
         static::assertFalse($page->hasMeta('property', 'foo'));
     }
 
+    /**
+     * @psalm-suppress InvalidArgument
```
jordisala1991, line 59: can we add a @see comment to explain why you have to supress?

jordisala1991 (changes requested):

tests/Seo/SeoPageTest.php
VincentLanglet: Not really, this error is made on purpose to trigger the exception in the test.
We want to check that if we provide an invalid argument, an exception is thrown.

VincentLanglet (dismissed):

src/DependencyInjection/SonataSeoExtension.php
```diff
@@ -52,6 +52,8 @@ public function load(array $configs, ContainerBuilder $container): void
 
     /**
      * Configure the default seo page.
+     *
+     * @param array<mixed> $config
```
jordisala1991: mixed[] (there are more uses like this).

src/Seo/SeoPage.php
```diff
@@ -272,7 +264,9 @@ public function getOEmbedLinks(): array
     }
 
     /**
-     * @param array|string $meta
+     * @param array<mixed>|string $meta
```
jordisala1991: string|mixed[]

src/DependencyInjection/Compiler/ServiceCompilerPass.php
```diff
@@ -33,10 +33,7 @@ public function process(ContainerBuilder $container): void
         $definition->addMethodCall('setHtmlAttributes', [$config['head']]);
         $definition->addMethodCall('setSeparator', [$config['separator']]);
 
-        if ($alias = $container->setAlias('sonata.seo.page', $config['default'])) {
-            $alias->setPublic(true);
-        }
-
+        $container->setAlias('sonata.seo.page', $config['default'])->setPublic(true);
```
franmomu, line 36: Not related to this PR, but do we need this alias to be public?

franmomu (approved):

jordisala1991 (approved):

VincentLanglet (approved):

PR #585, opened 2021-09-16
Drop support for PHP 7.3

core23:
## Subject

Drop support for PHP 7.3, because active support will end in a few days. This allows us to add type hints everywhere.

I am targeting this branch, because this is a major change.

## Changelog

```markdown
### Changed
- Added type hints to every property and method.

### Removed
- Dropped support for PHP 7.3

```

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -25,7 +25,6 @@ jobs:
         strategy:
             matrix:
                 php-version:
-                    - '7.3'
                     - '7.4'
                     - '8.0'
                 dependencies: [highest]
@@ -33,7 +32,7 @@ jobs:
                 symfony-require: [""]
                 variant: [normal]
                 include:
-                    - php-version: '7.3'
+                    - php-version: '7.4'
                       dependencies: lowest
                       allowed-to-fail: false
                       variant: normal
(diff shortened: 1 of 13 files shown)


.github/workflows/test.yaml
```diff
@@ -25,15 +25,14 @@ jobs:
         strategy:
             matrix:
                 php-version:
-                    - '7.3'
```
VincentLanglet, line 28: This should be on dev-kit

src/Event/BreadcrumbListener.php
```diff
@@ -24,10 +24,7 @@
  */
 final class BreadcrumbListener
 {
-    /**
-     * @var array
-     */
-    private $blockServices = [];
+    private array $blockServices = [];
```
VincentLanglet, line 27: We can still add phpdoc `array<...>`

VincentLanglet (approved):

src/Event/BreadcrumbListener.php
core23: Will check this when increasing phpstan / psalm level

jordisala1991:
Do you think it is a good idea for next majors to have different php required versions?

VincentLanglet:
> Do you think it is a good idea for next majors to have different php required versions?

I don't see any issue with this. You can still bump Php and then the major version.
Anyway, the major will be release potentially after the php version drop.

PR #588, opened 2021-09-16
Add `sonata_seo_breadcrumb` to render current breadcrumb

core23:
## Subject

Add new extension points to render the current breadcrumb using `sonata_seo_breadcrumb`.

The active breadcrumb can be set inside

I am targeting this branch, because this feature is BC.

## Changelog

```markdown
### Added
- Added `SeoPageInterface::setBreadcrumb`
- Added `SeoPageInterface::getBreadcrumbOptions`
- Added `sonata_seo_breadcrumb` twig extension to render current breadcrumb
```

### DIFF
--- a/src/Seo/SeoPage.php
+++ b/src/Seo/SeoPage.php
@@ -66,6 +66,11 @@ class SeoPage implements SeoPageInterface
     private $originalTitle;
 
     /**
+     * @var array<string, mixed>
+     */
+    private $breadcrumb;
+
+    /**
      * @param string $title
      */
     public function __construct($title = '')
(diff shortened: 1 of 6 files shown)


src/Seo/SeoPage.php
```diff
@@ -13,6 +13,8 @@
 
 namespace Sonata\SeoBundle\Seo;
 
+use Core23\SeoBundle\Seo\SeoPageInterface;
```
VincentLanglet: .

core23: 🤭

VincentLanglet (commented):
Should the breadcrumb be mandatory for a SeoPage ? Or should it be another interface ?

core23:
> Should the breadcrumb be mandatory for a SeoPage ? Or should it be another interface ?

Not sure if we could split the current `SeoPageInterface` into smaller ones. If you have a look at the [usage docs](https://github.com/sonata-project/SonataSeoBundle/blob/2.x/docs/reference/usage.rst), there is an example how you can chain the method calls. If you split them, you can't use the chaining anymore.

VincentLanglet (dismissed):

wbloszyk (dismissed):

VincentLanglet (approved):

VincentLanglet:
Thanks

PR #598, opened 2021-09-26
Fix null error in SeoPage

core23:
## Subject

There is a null error:

```
Return value of Sonata\SeoBundle\Seo\SeoPage::getBreadcrumbOptions() must be of the type array, null returned
```

I am targeting this branch, because this is a patch for an untagged release.

Fixes #588

### DIFF
--- a/src/Seo/SeoPage.php
+++ b/src/Seo/SeoPage.php
@@ -84,6 +84,7 @@ class SeoPage implements SeoPageInterface
             'property' => [],
         ];
 
+        // NEXT_MAJOR: Use property initialization
         $this->htmlAttributes = [];
         $this->headAttributes = [];
         $this->linkCanonical = '';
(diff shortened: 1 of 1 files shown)


src/Seo/SeoPage.php
```diff
@@ -68,7 +68,7 @@ class SeoPage implements SeoPageInterface
     /**
      * @var array<string, mixed>
      */
-    private $breadcrumb;
+    private $breadcrumb = [];
```
VincentLanglet, line 71: Can't we do the same for other property instead of doing it in the construct (for consistency and avoiding errors if someone override the construct)

core23: I've added a new note.

The other properties are `protected`, so we can't guarantee the default values.

VincentLanglet (approved):

PR #600, opened 2021-09-27
Allow array values for meta tags

### LINKED ISSUE #554: Allow array values for meta tags
## Feature Request

Currently if we want to add multiple _og:image_ tags, the **_addMeta_** function overrides the previous values and will use only the last one. According to the Open Graph specs at [https://ogp.me/#array](https://ogp.me/#array) it should be possible to define multiple tags with the same property name. SonataSeoBundle is currently not offering this feature as far as I'm aware of, but please correct me if I'm wrong.

BlackWiCKED:
## Subject

1. Updating **_addMeta()_** function to allow multiple values for specific meta tags: `og:image`

```
$page->addMeta('property', 'og:image', 'http://example.com/image1.jpg');
$page->addMeta('property', 'og:image', 'http://example.com/image2.jpg');
```
Renders as:
```
<meta property="og:image" content="http://example.com/image1.jpg">
<meta property="og:image" content="http://example.com/image2.jpg">
```

2. Utilize the **_$extra_** parameter to define structured properties when their key starts with `:`

```
$page->addMeta('property', 'og:image', 'http://example1.com/image1.jpg', [":width" => 640, ":height" => 480]);
```
Renders as:
```
<meta property="og:image" content="http://example.com/image1.jpg">
<meta property="og:image:width" content="640">
<meta property="og:image:height" content="480">
```

I am targeting this branch, because it's a missing feature and it should be available in 2.x too.

Closes #554

## Changelog

```markdown
### Added
- Extra parameters starting with a colon symbol (`:`) are now rendered as structured properties

### Changed
- Allow calling `SeoPage::addMeta()` with `og:image` multiple times
```

### DIFF
--- a/src/Seo/SeoPage.php
+++ b/src/Seo/SeoPage.php
@@ -21,6 +21,12 @@ namespace Sonata\SeoBundle\Seo;
 class SeoPage implements SeoPageInterface
 {
     /**
+     * @var array<string>
+     */
+    private const ALLOW_MULTIPLE_TAGS = [
+        'og:image',
+    ];
+    /**
      * @var string
      */
     protected $title;
@@ -157,7 +163,20 @@ class SeoPage implements SeoPageInterface
             $this->metas[$type] = [];
         }
 
-        $this->metas[$type][$name] = [$value, $extras];
+        $arrayValue = [$value, $extras];
+
+        if (\in_array($name, self::ALLOW_MULTIPLE_TAGS, true)) {
+            if (isset($this->metas[$type][$name])) {
+                [$oldValue, $oldExtras] = $this->metas[$type][$name];
+                if (!\is_array($oldValue)) {
+                    $oldValue = [$oldValue];
+                    $oldExtras = [$oldExtras];
+                }
+                $arrayValue = [array_merge($oldValue, [$value]), array_merge($oldExtras, [$extras])];
+            }
+        }
+
+        $this->metas[$type][$name] = $arrayValue;
 
         return $this;
     }
--- a/src/Seo/SeoPage.php
+++ b/src/Seo/SeoPage.php
@@ -21,7 +21,7 @@ namespace Sonata\SeoBundle\Seo;
 class SeoPage implements SeoPageInterface
 {
     /**
-     * @var array<string>
+     * @var string[]
      */
     private const ALLOW_MULTIPLE_TAGS = [
         'og:image',
--- a/src/Twig/Extension/SeoExtension.php
+++ b/src/Twig/Extension/SeoExtension.php
@@ -121,18 +121,27 @@ class SeoExtension extends AbstractExtension
                 [$content, $extras] = $meta;
 
                 if (!empty($content)) {
-                    $html .= sprintf(
-                        "<meta %s=\"%s\" content=\"%s\" />\n",
-                        $type,
-                        $this->normalize($name),
-                        $this->normalize($content)
-                    );
+                    if (\is_array($content)) {
+                        foreach ($content as $contentKey => $contentItem) {
+                            $html .= $this->generateMetaTag($type, $name, $contentItem);
+                            if (isset($extras[$contentKey]) && \is_array($extras[$contentKey])) {
+                                foreach ($extras[$contentKey] as $extraKey => $extra) {
+                                    if (':' === substr($extraKey, 0, 1)) {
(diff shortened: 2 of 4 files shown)


src/Seo/SeoPage.php
```diff
@@ -20,6 +20,12 @@
  */
 class SeoPage implements SeoPageInterface
 {
+    /**
+     * @var array<string>
```
core23:
```suggestion
     * @var string[]
```

src/Twig/Extension/SeoExtension.php
```diff
@@ -121,18 +121,27 @@ public function getMetadatas()
                 [$content, $extras] = $meta;
 
                 if (!empty($content)) {
-                    $html .= sprintf(
-                        "<meta %s=\"%s\" content=\"%s\" />\n",
-                        $type,
-                        $this->normalize($name),
-                        $this->normalize($content)
-                    );
+                    if (\is_array($content)) {
```
core23, line 124: This method has a large code complexity and is hard to understand. Can you try to extra a few methods.

VincentLanglet:
Wouldnt it better to say that the add method always has the same behavior to support array values for every possible tags ?

And if needed we can introduce later a reset/replace/set/... method.

jordisala1991:
> Wouldnt it better to say that the add method always has the same behavior to support array values for every possible tags ?
> 
> And if needed we can introduce later a reset/replace/set/... method.

I am not sure if all possible metatag allows multiple values. I guess not all of them do.

src/Twig/Extension/SeoExtension.php
BlackWiCKED: I can pass an _$extra_ variable to **_generateMetaTag()_** method as a new property, then the **_foreach_** would not be duplicated. The structured properties (additional tags) would be generated from the inside as a recursive function.
Or should I simply introduce a new **_generateMetaTagExtra()_** method and call it?

VincentLanglet:
Even if they dont, it would be better if you dont need to know if they do when you use the add method to determine the behavior you'll get.

Two different methods might be better

BlackWiCKED:
@VincentLanglet In my PR I introduced a constant (`ALLOW_MULTIPLE_TAGS`) where we can define which metatags shall accept array values. For now, I only added `og:image`, but I presume there are plenty more. Some of them are defined in https://ogp.me/#array but probably there are more meta tags (other than Open Graph) that allow multiple values. Just for the record, I don't support the idea of enabling this behavior for all tags. If that's what you aiming for, then this PR is obsolete, and there should be another one that stores the metadata in arrays by default all the time, and not just when we add more than one.

@core23 I understand the problem with the code complexity - and it's mainly because of the two options how we store the metadata. Most of the time it will be a string as it was before though - not much was added there. The real new feature is rendering structured data through the $extra variable - it can generate additional meta tags under the root node.

VincentLanglet (commented):
TLDR: This feature might seems easy but could introduce multiple inconsistency issues and BC-break to me. Might be simplier to think about a plan, add deprecation in this branch and finish it in next major. Good news the next major will be released soon.

Looking at the method:
- `setMetas` is resetting the metas and then calling addMeta with every value.
- `addMeta` is supposed to add a value, without replacing the previous one when the value is in this special array of "multiple values allowed". When it's not it will replace the value.
- a `setMeta` method could be introduce to reset and set a new value but it's basically `removeMeta` + `addMeta`.

The new strategy could add some issues:
- First adding/removing a key to this new constant is a BC-break ; because we're changing the behavior of the addMeta method.
- The addMeta won't have a consistent behavior depends of the tag used.
- The addMeta won't have a consistent behavior with all add method like `addHtmlAttributes`, `addHeadAttribute` (you can also notice the lack of consistency with the `s`).
- Also, you're checking the name but your usage is only in the case of a type `property`
- Since you're changing the structure of the metas, the call `setMetas(getMetas())` is not possible anymore. `foreach ($metaItem as $name => $meta) {` can give us array of array.

I don't like the `add` name to set a value. Renaming 
`addHtmlAttributes` to `setHtmlAttribute`, etc, would make sens.
then `removeHtmlAttributes` to `unsetHtmlAttribute`, etc, would make sens too.

The `addMeta` would be deprecated to a `setMeta` method which ALWAYS override the value.
And a new `addMeta` could be introduce and ALWAYS add the value to the existing array.

Also, it would be simplier to say that a meta is ALWAYS an array of array to avoid always checking if it's a string or an array and just doing a foreach on it.

PR #602, opened 2021-09-28
Move config from XML to PHP

wbloszyk:
Move configuration to PHP.


src/Resources/config/blocks.php
```diff
@@ -0,0 +1,35 @@
...
+use Sonata\SeoBundle\Block\Breadcrumb\HomepageBreadcrumbBlockService;
+use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
+use Symfony\Component\DependencyInjection\Loader\Configurator\ReferenceConfigurator;
+
+return static function (ContainerConfigurator $containerConfigurator): void {
+    // Use "service" function for creating references to services when dropping support for Symfony 4.4
+    // Use "param" function for creating references to parameters when dropping support for Symfony 5.1
+    $containerConfigurator->parameters()
+
+        ->set('sonata.seo.block.breadcrumb.homepage.class', HomepageBreadcrumbBlockService::class);
```
core23, line 23: We should remove all class parameters

wbloszyk: I just thinking. Mybe we should also do it in `2.x`?

core23: IMHO it's a breaking change if we remove this on the stable branch.

wbloszyk: Move config to PHP in `2.x` and remove parameters in `3.x`.

core23:
Done with #604

PR #606, opened 2021-09-29
Change SourceManager implementation

core23:
## Subject

Fixes the build by changing the `SourceManager` implementation. The class has nothing to do with the exporter, so there is no need to implement the interface.

I am targeting this branch, because this is a breaking change.

## Changelog

```markdown
### Changed
- Change `SourceManager` implementation
```

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1,9 +1,2 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<files psalm-version="4.10.0@916b098b008f6de4543892b1e0651c1c3b92cbfa">
-    <file src="src/Sitemap/SourceManager.php">
-        <InvalidTemplateParam occurrences="2">
-            <code>SourceIteratorInterface</code>
-            <code>SourceIteratorInterface&lt;\stdClass&gt;</code>
-        </InvalidTemplateParam>
-    </file>
-</files>
+<files psalm-version="4.10.0@916b098b008f6de4543892b1e0651c1c3b92cbfa"/>
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1,2 +1,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<files psalm-version="4.10.0@916b098b008f6de4543892b1e0651c1c3b92cbfa"/>
--- a/src/Sitemap/Source.php
+++ b/src/Sitemap/Source.php
@@ -1,0 +1,58 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\SeoBundle\Sitemap;
+
+use Sonata\Exporter\Source\ChainSourceIterator;
+use Sonata\Exporter\Source\SourceIteratorInterface;
+
+final class Source
+{
+    /**
+     * @var mixed[]
+     */
+    private array $types = [];
+
+    private ChainSourceIterator $sources;
+
+    public function __construct()
+    {
+        $this->sources = new ChainSourceIterator();
+    }
+
+    public function getSources(): ChainSourceIterator
+    {
+        return $this->sources;
+    }
+
+    public function addSource(SourceIteratorInterface $source): void
+    {
+        $this->sources->addSource($source);
+    }
+
+    /**
+     * @return mixed[]
+     */
+    public function getTypes(): array
+    {
+        return $this->types;
+    }
+
+    /**
+     * @param mixed[] $types
+     */
+    public function addTypes(array $types): void
+    {
+        $this->types += array_diff($types, $this->types);
+    }
+}
--- a/src/Sitemap/SourceManager.php
+++ b/src/Sitemap/SourceManager.php
@@ -19,9 +19,9 @@ use Sonata\Exporter\Source\SourceIteratorInterface;
 /**
  * Manager several chain source iterator grouped.
  *
- * @phpstan-implements SourceIteratorInterface<\stdClass>
+ * @phpstan-implements \Iterator<\stdClass>
  */
-final class SourceManager implements SourceIteratorInterface
+final class SourceManager implements \Iterator
 {
     /**
      * @var \ArrayIterator<string, \stdClass>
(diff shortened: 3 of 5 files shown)


src/Sitemap/SourceManager.php
```diff
@@ -19,9 +19,9 @@
 /**
  * Manager several chain source iterator grouped.
  *
- * @phpstan-implements SourceIteratorInterface<\stdClass>
+ * @phpstan-implements \Iterator<\stdClass>
```
VincentLanglet: Shouldn't we use a ValueObject instead of a stdClass ?

VincentLanglet: Another option could be to use an array ['source' => ..., 'type' => ...] ; this way it respects the interface template.

src/Sitemap/SourceManager.php
```diff
@@ -36,8 +36,7 @@ public function __construct()
     /**
      * Adding source with his group.
      *
-     * @param SourceIteratorInterface<\stdClass> $source
-     * @param mixed[]                            $types
+     * @param mixed[] $types
      */
     public function addSource(string $group, SourceIteratorInterface $source, array $types = []): void
```
VincentLanglet, line 40: This class is not _supposed_ to be related to the exporter, but the interface is still used. What's the idea behind ?

Also when I see the interface SourceIteratorInterface, https://github.com/sonata-project/exporter/blob/2.x/src/Source/SourceIteratorInterface.php I kinda wonder if we shouldn't just deprecate it in favor of \Iterator.

src/Sitemap/SourceManager.php
core23: Done

src/Sitemap/SourceManager.php
core23: > This class is not _supposed_ to be related to the exporter, but the interface is still used. What's the idea behind ?

That's not true. The class contains many `SourceIteratorInterface` which are used in the `SitemapGeneratorCommand` to generate the sitemap.

> Also when I see the interface SourceIteratorInterface, https://github.com/sonata-project/exporter/blob/2.x/src/Source/SourceIteratorInterface.php I kinda wonder if we shouldn't just deprecate it in favor of \Iterator.

👍

src/Sitemap/Source.php
```diff
@@ -0,0 +1,31 @@
...
+use Sonata\Exporter\Source\ChainSourceIterator;
+
+final class Source
+{
+    /**
+     * @var mixed[]
+     */
+    public array $types = [];
+
+    public ChainSourceIterator $sources;
```
VincentLanglet: Dunno if it's better to use a private property with getter/setter/adder/...

VincentLanglet (dismissed):

src/Sitemap/Source.php
core23: Done

src/Sitemap/SourceManager.php
```diff
@@ -13,18 +13,17 @@
...
 namespace Sonata\SeoBundle\Sitemap;
 
-use Sonata\Exporter\Source\ChainSourceIterator;
 use Sonata\Exporter\Source\SourceIteratorInterface;
 
 /**
  * Manager several chain source iterator grouped.
  *
- * @phpstan-implements SourceIteratorInterface<\stdClass>
+ * @phpstan-implements \Iterator<Source>
```
franmomu:
```suggestion
 * @phpstan-implements \Iterator<string, Source>
```

psalm-baseline.xml
```diff
@@ -1,9 +1,2 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<files psalm-version="4.10.0@916b098b008f6de4543892b1e0651c1c3b92cbfa">
-    <file src="src/Sitemap/SourceManager.php">
-        <InvalidTemplateParam occurrences="2">
-            <code>SourceIteratorInterface</code>
-            <code>SourceIteratorInterface&lt;\stdClass&gt;</code>
-        </InvalidTemplateParam>
-    </file>
-</files>
+<files psalm-version="4.10.0@916b098b008f6de4543892b1e0651c1c3b92cbfa"/>
```
franmomu: We could remove this file if it's empty

src/Sitemap/Source.php
```diff
@@ -0,0 +1,58 @@
...
+     */
+    public function getTypes(): array
+    {
+        return $this->types;
+    }
+
+    /**
+     * @param mixed[] $types
+     */
+    public function addType(array $types): void
```
VincentLanglet: addTypes maybe ?

Or the addType should just add one type, and we should have the array_diff logic somewhere else.

Also the only usage of `addSource` is here: https://github.com/sonata-project/SonataSeoBundle/blob/f4ab958ccd91e56f3e3cb21d156a08bfb85a9a71/src/DependencyInjection/SonataSeoExtension.php#L99
Can't we know what is a type to give a more specific typehint than mixed ? (never used this bundle but I assume it's a string in https://docs.sonata-project.org/projects/SonataSeoBundle/en/3.x/reference/sitemap/#configuration-example)

Then having unique values are kinda easy with strings...

VincentLanglet (commented):
Doc need to be updated too https://docs.sonata-project.org/projects/SonataSeoBundle/en/3.x/reference/sitemap/#service-definition

core23:
> Doc need to be updated too https://docs.sonata-project.org/projects/SonataSeoBundle/en/3.x/reference/sitemap/#service-definition

Why? Only the container for all `SourceIteratorInterface` was refactored. 

The docs need to be updated when using the next exporter major version which will remove the `SourceIteratorInterface`.

VincentLanglet:
> > Doc need to be updated too https://docs.sonata-project.org/projects/SonataSeoBundle/en/3.x/reference/sitemap/#service-definition
> 
> 
> 
> Why? Only the container for all `SourceIteratorInterface` was refactored. 
> 
> 
> 
> The docs need to be updated when using the next exporter major version which will remove the `SourceIteratorInterface`.

My bad. I thought it wasnt needed to implement the interface and Iterator was only needed. Indeed it will only be changed in the next major.

src/Sitemap/Source.php
core23: > addTypes maybe ?
> 

Renamed the method.

> Also the only usage of `addSource` is here:
> 
> https://github.com/sonata-project/SonataSeoBundle/blob/f4ab958ccd91e56f3e3cb21d156a08bfb85a9a71/src/DependencyInjection/SonataSeoExtension.php#L99

That's not correct. The `addSource` method here calls the method on `SourceIteratorInterface`.

> Can't we know what is a type to give a more specific typehint than mixed ? (never used this bundle but I assume it's a string in https://docs.sonata-project.org/projects/SonataSeoBundle/en/3.x/reference/sitemap/#configuration-example)

Never used this part either. For what I know, the `type` is just an array with an unknown signature. I would keep it as it is. 

This PR started to just fix the broken `3.x` build.

core23:
Please review @sonata-project/contributors so we can fix the broken `3.x` branch. I would like to release a new major version soon.

VincentLanglet (approved):

franmomu (approved):

PR #608, opened 2021-10-06
Merge 2.x into 3.x

core23:


franmomu (dismissed):

src/Resources/config/blocks.php
```diff
@@ -0,0 +1,38 @@
...
+            ->public()
+            ->tag('sonata.block')
+            ->tag('sonata.breadcrumb')
+            ->args([
+                'homepage',
+                'sonata.seo.block.breadcrumb.homepage',
+                new ReferenceConfigurator('sonata.templating'),
+                new ReferenceConfigurator('knp_menu.menu_provider'),
+                new ReferenceConfigurator('knp_menu.factory'),
+            ]);
```
wbloszyk: Can you set correct arguments here?

src/BreadcrumbInterface.php
```diff
@@ -15,6 +15,9 @@
 
 use Sonata\BlockBundle\Block\Service\BlockServiceInterface;
 
+/**
+ * @deprecated since sonata-project/seo-bundle 2.x, to be removed in 3.0. Use Sonata\SeoBundle\Block\Breadcrumb\Breadcrumb instead.
```
wbloszyk, line 19: IMO should be:
```suggestion
 * @deprecated since sonata-project/seo-bundle 2.x, to be removed in 3.0. Use Sonata\SeoBundle\Block\Breadcrumb\BreadcrumbBlockService instead.
```

wbloszyk (changes requested):

src/BreadcrumbInterface.php
VincentLanglet: This is just a merge, such a change should be done on 2.x branch to avoid conflict

src/Resources/config/blocks.php
VincentLanglet: What is wrong with the arguments ?

wbloszyk: Look at https://github.com/sonata-project/SonataSeoBundle/pull/608/files/f6d0606ac271a593e090363c5b5a5e45a9f76cd4#diff-953b23c1f0b95f9eb30962f3590b811ce55cb4a0191a2417a1f2f90adbae9c95

VincentLanglet (approved):

wbloszyk (approved):

PR #609, opened 2021-10-07
Prepare major release

core23:
## Subject

I am targeting this branch, because this should be the last changes for the next major release.

## Changelog

```markdown
### Changed
- Made `sonata.seo.sitemap.manager` service private

### Removed
- Remove all `sonata.seo.*.class` parameters from container
```


VincentLanglet:
If we follow the same strategy than other bundle, we should also 
- Bump phpstan level 8: https://github.com/sonata-project/SonataSeoBundle/blob/3.x/phpstan.neon.dist#L5
- Add the strict-rules plugin of phpstan (we found a lot of bug with it on SonataAdmin)
- Bump psalm level 3: https://github.com/sonata-project/SonataSeoBundle/blob/3.x/psalm.xml#L2

And we might determine what we do with https://github.com/sonata-project/SonataSeoBundle/pull/600#pullrequestreview-766353269

core23:
> * Bump phpstan level 8: https://github.com/sonata-project/SonataSeoBundle/blob/3.x/phpstan.neon.dist#L5
> * Add the strict-rules plugin of phpstan (we found a lot of bug with it on SonataAdmin)
> * Bump psalm level 3: https://github.com/sonata-project/SonataSeoBundle/blob/3.x/psalm.xml#L2

Done

> And we might determine what we do with [#600 (review)](https://github.com/sonata-project/SonataSeoBundle/pull/600#pullrequestreview-766353269)

We can still implement the feature on the next major version

src/Resources/config/services.php
```diff
@@ -22,30 +20,18 @@
...
-        ->set('sonata.seo.sitemap.manager.class', SourceManager::class);
-
     $containerConfigurator->services()
 
-        ->set('sonata.seo.page.default', '%sonata.seo.page.default.class%')
+        ->set('sonata.seo.page.default', SeoPage::class)
             ->public()
 
-        ->set('sonata.seo.twig.extension', '%sonata.seo.twig.extension.class%')
+        ->set(SeoExtension::class)
```
franmomu: We are using prefixed strings for the service names as suggested by Symfony: https://symfony.com/doc/current/bundles/best_practices.html#services

src/Resources/config/services.php
```diff
@@ -22,30 +20,18 @@
...
-
-        ->set('sonata.seo.twig.extension.class', SeoExtension::class)
-
-        ->set('sonata.seo.sitemap.manager.class', SourceManager::class);
-
     $containerConfigurator->services()
 
-        ->set('sonata.seo.page.default', '%sonata.seo.page.default.class%')
+        ->set('sonata.seo.page.default', SeoPage::class)
             ->public()
```
franmomu, line 26: I haven't checked the usages, but do we need services to be public?

src/Sitemap/SourceManager.php
```diff
@@ -39,10 +39,12 @@ public function __construct()
      */
     public function addSource(string $group, SourceIteratorInterface $source, array $types = []): void
     {
-        if (!isset($this->sources[$group])) {
+        if (!isset($this->sources[$group]) || null === $this->sources[$group]) {
```
franmomu: is the right branch needed? if `$this->sources[$group]` is null, `!isset($this->sources[$group])` would be `true` as well.

src/Resources/config/services.php
core23: Reverted this. 

Looks like every symfony extension has a named service id: https://github.com/symfony/symfony/blob/ed712ef218a66d0bb1f659e62b4be1a47ab4c14a/src/Symfony/Bundle/TwigBundle/Resources/config/twig.php

src/Resources/config/services.php
core23: Not sure if symfony still recommends using the service container. If it's deprecated we could make this private.

franmomu: [Looks like it depends on the usage](https://symfony.com/doc/5.2/bundles/best_practices.html#services):

> In addition, services not meant to be used by the application directly, should be defined as private. For public services, aliases should be created from the interface/class to the service id.

(I haven checked, but I guess they don't recommend using the service container directly.)

franmomu (approved):

src/Block/Breadcrumb/BreadcrumbBlockService.php
```diff
@@ -13,8 +13,9 @@
 
 namespace Sonata\SeoBundle\Block\Breadcrumb;
 
-use Sonata\SeoBundle\BreadcrumbInterface;
+use Sonata\BlockBundle\Block\Service\BlockServiceInterface;
```
wbloszyk, line 16: I dont see this interface.

src/DependencyInjection/SonataSeoExtension.php
```diff
@@ -75,7 +78,7 @@ private function configureSitemap(array $config, ContainerBuilder $container): v
             // define the connectionIterator
             $connectionIteratorId = 'sonata.seo.source.doctrine_connection_iterator_'.$pos;
 
-            $connectionIterator = new Definition('%sonata.seo.exporter.database_source_iterator.class%', [
+            $connectionIterator = new Definition(DoctrineDBALConnectionSourceIterator::class, [
```
wbloszyk, line 81: I'm not familiar with this, but if we have paramteter here should not be better to keep it and set it in config?

src/DependencyInjection/SonataSeoExtension.php
```diff
@@ -86,7 +89,7 @@ private function configureSitemap(array $config, ContainerBuilder $container): v
             // define the sitemap proxy iterator
             $sitemapIteratorId = 'sonata.seo.source.doctrine_sitemap_iterator_'.$pos;
 
-            $sitemapIterator = new Definition('%sonata.seo.exporter.sitemap_source_iterator.class%', [
+            $sitemapIterator = new Definition(SymfonySitemapSourceIterator::class, [
```
wbloszyk, line 92: I'm not familiar with this, but if we have paramteter here should not be better to keep it and set it in config?

wbloszyk (changes requested):

wbloszyk:
Before next major release we should decide about one importan thing.

After 2 months (59 days) support for PHP 7.3 will be end. If we drop it now then we can also declare types for parameters. Do it in stable branch will be BC. 

WDYT? @sonata-project/contributors

src/Block/Breadcrumb/BreadcrumbBlockService.php
wbloszyk: My bad, not this bundle :)

franmomu:
> Before next major release we should decide about one importan thing.
> 
> After 2 months (59 days) support for PHP 7.3 will be end. If we drop it now then we can also declare types for parameters. Do it in stable branch will be BC.
> 
> WDYT? @sonata-project/contributors

It's already dropped in this bundle: https://github.com/sonata-project/SonataSeoBundle/blob/23cf25df16033c5197d2d0312118b175e9fdb9cf/composer.json#L23

wbloszyk:
I want check it this bundle before major release in the evening. So can we wait with major release to tomorrow?

VincentLanglet (approved):

src/DependencyInjection/SonataSeoExtension.php
core23: We could improve this by allowing a custom service via the config, but for now we should get rid of all class parameters.

wbloszyk (approved):

VincentLanglet:
> I want check it this bundle before major release in the evening. So can we wait with major release to tomorrow?

We should try with an alpha/beta version anyway

PR #621, opened 2021-11-01
Prepare 3.0.0-RC1

core23:
As there is not so much activity around this project, it should be safe to create the first release candidate version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,21 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [3.0.0-RC1](https://github.com/sonata-project/SonataSeoBundle/compare/3.0.0-alpha.1...3.0.0-RC1) - 2021-11-03
+### Added
+- [[#614](https://github.com/sonata-project/SonataSeoBundle/pull/614)] Symfony 6 support ([@Kocal](https://github.com/Kocal))
+
+### Changed
+- [[#624](https://github.com/sonata-project/SonataSeoBundle/pull/624)] `BaseBreadcrumbMenuBlockService::getMenu()` method contains the root menu ([@core23](https://github.com/core23))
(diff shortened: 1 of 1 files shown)


VincentLanglet (dismissed):
Phpstan level 8 with strict rules
Psalm level 3
No next major
typehint
final classes (Maybe we could add final to 
https://github.com/sonata-project/SonataSeoBundle/blob/9f3287a72ca2fecb81708f4dc3fa2c69eb474ac1/src/Block/Breadcrumb/BaseBreadcrumbMenuBlockService.php#L40
too ?)

seems ok to me

core23:
> inal classes (Maybe we could add final to
> 
> https://github.com/sonata-project/SonataSeoBundle/blob/9f3287a72ca2fecb81708f4dc3fa2c69eb474ac1/src/Block/Breadcrumb/BaseBreadcrumbMenuBlockService.php#L40
> 
> 
> too ?)

Good catch. We don't need this method. I'll fix this with #622

jordisala1991:
Maybe we can avoid rc and go directly for 3.0? If there is no reported issues for the alpha and barely no changes

VincentLanglet:
> > inal classes (Maybe we could add final to
> > https://github.com/sonata-project/SonataSeoBundle/blob/9f3287a72ca2fecb81708f4dc3fa2c69eb474ac1/src/Block/Breadcrumb/BaseBreadcrumbMenuBlockService.php#L40
> > 
> > too ?)
> 
> Good catch. We don't need this method. I'll fix this with #622

Also, HomepageBreadcrumbBlockService is overriding `getMenu` to do the same implementation.
Can't `getMenu` become private ? (and I see no reason to create another method `getRootMenu` then)

If we decide to make `getMenu` non-final and protected to allow an override, then `getMenuOptions` should be private too. (But still I see no reason to have a `getRootMenu` method)

core23:
Please review @sonata-project/contributors. I'll plan to release this tomorrow

VincentLanglet (approved):

wbloszyk (approved):

PR #636, opened 2021-11-18
Drop support for PHP 7

core23:
## Subject

Active support will end in a few days: https://www.php.net/supported-versions.php

We have the chance to remove all PHP 7 related code before releasing the final 3.0 release.

I am targeting this branch, because we can do this for the unstable branch.

## Changelog

```markdown
### Removed
- PHP 7 support
```

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -26,33 +26,33 @@ jobs:
         strategy:
             matrix:
                 php-version:
-                    - '7.4'
                     - '8.0'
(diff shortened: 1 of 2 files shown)


VincentLanglet (changes requested):
Imho since we released a RC, we should follow the same strategy than a stable branch and keep support for php version with security fixes.


We just dropped 7.3 ; this seems enough.

And I see no benefit for now except blocking users.

PR #641, opened 2021-11-28
DevKit updates for 3.x branch

SonataCI:


jordisala1991 (approved):

src/Sitemap/SourceManager.php
```diff
@@ -52,6 +52,12 @@ public function addSource(string $group, SourceIteratorInterface $source, array
         }
     }
 
+    /**
+     * @return mixed
+     *
+     * @phpstan-return Source
```
VincentLanglet, line 58: Why not just `return Source` ?

src/Sitemap/SourceManager.php
```diff
@@ -62,6 +68,12 @@ public function next(): void
         $this->sources->next();
     }
 
+    /**
+     * @return mixed
+     *
+     * @phpstan-return string
```
VincentLanglet, line 74: why not just return string ?

VincentLanglet (changes requested):

src/Sitemap/SourceManager.php
jordisala1991: the parent class is returning mixed, on next major I guess we should add `: mixed`

src/Sitemap/SourceManager.php
jordisala1991: Same as above, parent class is returning mixed. We should not forget to add the correct return type

src/Sitemap/SourceManager.php
VincentLanglet: This seems to not be an issue, see https://github.com/sonata-project/SonataSeoBundle/pull/642

PR #642, opened 2021-11-28
Avoid phpstan-return

VincentLanglet:


jordisala1991:
Did you read my last comment on the previous Pr?

VincentLanglet:
> Did you read my last comment on the previous Pr?

Test are fine, so I don't see why it's an issue to have a more precise return type.
In the futur we will have both `return Source` and `: mixed`, that's not a problem

core23 (approved):

jordisala1991:
I dont see how we will remember the correct type, but it is fine by me

VincentLanglet:
> I dont see how we will remember the correct type, but it is fine by me

I'm not even sure it's important to add `: mixed`.
Since php 7.4, a more precise return typehint can be used, so 
```
public function current(): Source
```
should be valid.

jordisala1991 (approved):

Issue #520, opened 2021-03-05
[RFC] Rethink breadcrumbs


core23:
## Feature Request

There are some issues with the breadcrumb component (#446, #443). They are caused by a design issue that was introduced a long time ago.

At the moment a breadcrumb MUST extend the `BaseBreadcrumbMenuBlockService` which is a **breadcrumb** AND a **block**. This component is then render via https://github.com/sonata-project/SonataBlockBundle/blob/3.x/src/Block/Service/MenuBlockService.php. So a block renders an other block oO

A better idea is to split this component into a breadcrumb and a block which renders a breadcrumb.

```php
namespace Sonata\SeoBundle\Breadcrumb;

interface Breadcrumb 
{
    public function configureSettings(OptionsResolver $resolver): void;

    public function getMenu(array $options = []): ItemInterface;
}
```

wbloszyk:
If I undersand correctly for now it is working as block: 
https://github.com/sonata-project/ecommerce/blob/3.x/src/ProductBundle/Block/CatalogBreadcrumbBlockService.php
Render breadcrumb it possible by something like:
`{{ sonata_block_render('sonata.product.block.breadcrumb') }}`
or
`{{ sonata_block_render_event('breadcrumb', { 'context': 'news_post', 'post': post, 'current_uri': app.request.requestUri }) }}`

IMO we should change for example `CatalogBreadcrumbBlockService` to `CatalogBreadcrumbService` and render it by twig function. Additionaly for use it by block we can add final `BreadcrumbBlockService` which will render `CatalogBreadcrumbService` object.

Also we should allow use advance build menu and display only this `ItemInterface` where `isCurrent()` return `true`. IMO this case can be render by `MenuBlockService` with setting - `display_currents_only` = true.

Issue #446, opened 2020-10-11
BaseBreadcrumbMenuBlockService is broken in master


greg0ire:
In https://github.com/sonata-project/SonataSeoBundle/pull/399 (a PR made against master), `BaseBreadcrumbMenuBlockService` parent was changed from `Sonata\BlockBundle\Block\Service\MenuBlockService` to `Sonata\BlockBundle\Block\Service\AbstractBlockService`. This results in lost functionality as revealed by a test I wrote in https://github.com/sonata-project/SonataSeoBundle/pull/426

More details at https://github.com/sonata-project/SonataSeoBundle/commit/84355f6ed55d7a3c95dcfdf2bbfbc1f6ee385eee#r42221263

Might be related to #443 too.

core23:
Closing this in favor of #520

Issue #443, opened 2020-10-08
BaseBreadcrumbMenuBlockService should not extends MenuBlockService


VincentLanglet:
Then we can make the Sonata\BlockBundle\Block\Service\MenuBlockService class final again.

https://github.com/sonata-project/SonataBlockBundle/pull/761

core23:
Maybe we should completely rethink this component. A breadcrumb should have nothing to do with a BlockService.

My idea is something like this:

```php
namespace Sonata\SeoBundle\Breadcrumb;

interface Breadcrumb {
    public function configureSettings(OptionsResolver $resolver): void;

    public function getMenu(array $options = []): ItemInterface;
}


```

VincentLanglet:
There is already a BreadcrumbInterface. https://github.com/sonata-project/SonataSeoBundle/blob/a3a55dcaea4d3acd9284e842a09fe4b566b94cda/src/BreadcrumbInterface.php

If we remove the `extends MenuBlockService` and add some methods, it could be enough.

I have some trouble to find why we extending `MenuBlockService` indeed. But I never used this bundle.
Do you mind creating a PR/a draft @core23 ? :)

core23:
> I have some trouble to find why we extending `MenuBlockService` indeed.

It has happend a long time ago when extend as many classes as possible was *in vogue* ;)

> Do you mind creating a PR/a draft @core23 ? :)

Sadly I haven't much time for this huge refactoring. We also need some adjustments in the `MenuBlock` to get this thing working.

greg0ire:
> I have some trouble to find why we extending MenuBlockService indeed. But I never used this bundle.

I believe we did so to benefit from this piece of code: https://github.com/sonata-project/SonataSeoBundle/commit/84355f6ed55d7a3c95dcfdf2bbfbc1f6ee385eee#r42421762

wbloszyk:
We can close it and fix `BaseBreadcrumbMenuBlockService` in #446

PR #356, opened 2019-11-29
Remove old sonata dependencies

core23:
## Subject

Make optional dependency for sonata block a required dependency.

I am targeting this branch, because this is not BC.

## Changelog

```markdown
### Added
- Added `sonata-project/block-bundle` dependency

### Changed
- Adopt new block service interfaces
```

## To do
- [x] New form release    
- [ ] Update the tests
- [ ] Create `MenuBlock` interface in BlockBundle

PR #691, opened 2022-09-15
Revert bc break

core23:
## Subject

I am targeting this branch, because this reverts the BC break introduced with #687 

## Changelog

```markdown
### Added
- Added `EditableBreadcrumbMenuBlockService`

### Fixed
- [BC break] Breadcrumbs now work without SonataPageBundle blocks again, broken since 3.3
```


src/Block/Breadcrumb/EditableBreadcrumbMenuBlockService.php
```diff
@@ -0,0 +1,85 @@
...
+use Sonata\BlockBundle\Meta\MetadataInterface;
+use Sonata\BlockBundle\Model\BlockInterface;
+use Sonata\Form\Validator\ErrorElement;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+use Twig\Environment;
+
+/**
+ * @author Sylvain Deloux <sylvain.deloux@ekino.com>
+ */
+abstract class EditableBreadcrumbMenuBlockService extends BaseBreadcrumbMenuBlockService implements EditableBlockService
```
core23, line 26: This class should not be here, but it looks like it is used by some people...

src/Block/Breadcrumb/BaseBreadcrumbMenuBlockService.php
```diff
@@ -17,96 +17,64 @@
...
-use Sonata\BlockBundle\Form\Mapper\FormMapper;
-use Sonata\BlockBundle\Meta\MetadataInterface;
-use Sonata\BlockBundle\Model\BlockInterface;
-use Sonata\Form\Validator\ErrorElement;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\OptionsResolver\OptionsResolver;
 use Twig\Environment;
 
 /**
+ * Abstract class for breadcrumb menu services.
```
VincentLanglet: Not really useful comment.

src/Block/Breadcrumb/EditableBreadcrumbMenuBlockService.php
```diff
@@ -0,0 +1,85 @@
...
+use Sonata\BlockBundle\Block\Service\EditableBlockService;
+use Sonata\BlockBundle\Form\Mapper\FormMapper;
+use Sonata\BlockBundle\Meta\MetadataInterface;
+use Sonata\BlockBundle\Model\BlockInterface;
+use Sonata\Form\Validator\ErrorElement;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+use Twig\Environment;
+
+/**
+ * @author Sylvain Deloux <sylvain.deloux@ekino.com>
```
VincentLanglet: Wrong author.

src/Block/Breadcrumb/EditableBreadcrumbMenuBlockService.php
VincentLanglet: Why shouldn't it be here ?
What if two different bundles are using it ?

src/Block/Breadcrumb/EditableBreadcrumbMenuBlockService.php
```diff
@@ -0,0 +1,85 @@
...
+ */
+abstract class EditableBreadcrumbMenuBlockService extends BaseBreadcrumbMenuBlockService implements EditableBlockService
+{
+    /**
+     * @var BlockServiceInterface&EditableBlockService
+     */
+    private object $menuBlock;
```
VincentLanglet:
```suggestion
    private BlockServiceInterface $menuBlock;
```

src/Block/Breadcrumb/EditableBreadcrumbMenuBlockService.php
```diff
@@ -0,0 +1,85 @@
...
+
+    /**
+     * @param BlockServiceInterface&EditableBlockService $menuBlock
+     */
+    public function __construct(Environment $twig, object $menuBlock, FactoryInterface $factory)
```
VincentLanglet:
```suggestion
    public function __construct(Environment $twig, BlockServiceInterface $menuBlock, FactoryInterface $factory)
```

src/Block/Breadcrumb/EditableBreadcrumbMenuBlockService.php
```diff
@@ -0,0 +1,85 @@
...
+    {
+        // TODO: Remove this check and add the intersection typehint when dropping support for PHP < 8.1
+        if (!$menuBlock instanceof BlockServiceInterface || !$menuBlock instanceof EditableBlockService) {
```
VincentLanglet:
```suggestion
        if (!$menuBlock instanceof EditableBlockService) {
```

src/Resources/config/blocks.php
```diff
@@ -27,7 +27,6 @@
             ->tag('sonata.breadcrumb')
             ->args([
                 new ReferenceConfigurator('twig'),
-                new ReferenceConfigurator('sonata.block.service.menu'),
```
VincentLanglet, line 30: So the HomepageBreadcrumbBlockService doesn't need to be an editableBlockService ?

src/Block/Breadcrumb/BaseBreadcrumbMenuBlockService.php
```diff
@@ -17,96 +17,64 @@
...
-        $this->menuBlock->validate($errorElement, $block);
+        return $this->renderResponse($template, $responseSettings, $response);
     }
 
     public function configureSettings(OptionsResolver $resolver): void
     {
-        $this->menuBlock->configureSettings($resolver);
-
         $resolver->setDefaults([
+            'cache_policy' => 'public',
```
VincentLanglet, line 61: We will need to deprecate this option since it's deprecated in BlockBundle and renderPrivateResponse method won't exist.

src/Block/Breadcrumb/EditableBreadcrumbMenuBlockService.php
```diff
@@ -0,0 +1,85 @@
...
+
+    public function configureSettings(OptionsResolver $resolver): void
+    {
+        $this->menuBlock->configureSettings($resolver);
+
+        $resolver->setDefaults([
+            'menu_template' => '@SonataSeo/Block/breadcrumb.html.twig',
+            'include_homepage_link' => true,
+            'context' => null,
+        ]);
```
VincentLanglet, line 80: This part is already in `BaseBreadcrumbMenuBlockService::configureSettings`, shouldn't we do 
```
parent::configureSettings($resolver);
```
instead ?

tests/Block/Breadcrumb/BreadcrumbTest.php
```diff
@@ -16,18 +16,12 @@
...
 use Sonata\BlockBundle\Block\BlockContext;
-use Sonata\BlockBundle\Block\Service\BlockServiceInterface;
-use Sonata\BlockBundle\Block\Service\EditableBlockService;
 use Sonata\BlockBundle\Model\Block;
 use Sonata\BlockBundle\Test\BlockServiceTestCase;
 use Sonata\SeoBundle\Block\Breadcrumb\BaseBreadcrumbMenuBlockService;
 use Symfony\Component\HttpFoundation\Response;
 use Twig\Environment;
 
-interface EditableBlockServiceInterface extends BlockServiceInterface, EditableBlockService
```
VincentLanglet, line 27: We need an EditableBreadcrumbTest then.

VincentLanglet (commented):
It's not clear to me what's the difference between a BlockService and an EditableBlockService. And then, it's not clear to me what's the benefit of this PR.

People already started using the 3.3.0 version (https://packagist.org/packages/sonata-project/seo-bundle/stats#3.3.0), and we should keep in mind that reverting this BC-break will also be a BC-break.
So if it's just to revert the BC-break, it might be not worth it imho.
If there is another benefit, it could be nice to explain it.

Seems like the doc is not up to date https://github.com/sonata-project/SonataBlockBundle/blob/4.x/docs/reference/your_first_block.rst since it explains that all the block need a `buildEditForm` and `validateBlock` method.

Maybe it would be helpful to update this doc, in order to understand what is the benefit of providing a blockservice which is not an editable block service.

src/Block/Breadcrumb/EditableBreadcrumbMenuBlockService.php
core23: It should be part of the SonataPageBundle, because it is only used there

core23:
> It's not clear to me what's the difference between a BlockService and an EditableBlockService. And then, it's not clear to me what's the benefit of this PR.

YAGNI. The general idea is to support blocks without any editable part. So you don't need all dependencies for a simple breadcrumb.

> People already started using the 3.3.0 version (https://packagist.org/packages/sonata-project/seo-bundle/stats#3.3.0), and we should keep in mind that reverting this BC-break will also be a BC-break. So if it's just to revert the BC-break, it might be not worth it imho. If there is another benefit, it could be nice to explain it.

... and I am using the 3.0 block version for many months without any problems which is a problem since a few days. Reverting a newly introduced BC break should be no general BC break.

> Seems like the doc is not up to date https://github.com/sonata-project/SonataBlockBundle/blob/4.x/docs/reference/your_first_block.rst since it explains that all the block need a `buildEditForm` and `validateBlock` method.
> 
> Maybe it would be helpful to update this doc, in order to understand what is the benefit of providing a blockservice which is not an editable block service.

This project has nothing to do with the SonataPageBundle. So we should keep this in mind when working on a standalone lib. As mentioned, the editable part should be moved to the SonataPageBundle.

jordisala1991:
> This project has nothing to do with the SonataPageBundle. So we should keep this in mind when working on a standalone lib. As mentioned, the editable part should be moved to the SonataPageBundle.

And it is still the case without your PR. This bundle depends on SonataBlockBundle which offers a EditableBlockService interface to make your blocks editable, and it still doesn't mean anything of SonataPageBundle. You can edit your blocks with normal symfony forms building your own flow.

> YAGNI. The general idea is to support blocks without any editable part. So you don't need all dependencies for a simple breadcrumb.

YAGNI doesn't mean a lot on a lib, your needs are not my needs, why should we prioritize yours then? The point mentioned in several issues by you is: Breadcrumb shouldn't be a block in the first place. I can agree with that, didn't take a deeper look at the code to see how possible is that. But in 3.0.0 be broke BC in a way that something that was previously possible (edit breadcrumbs) is now not possible, without giving an alternative, and with the thinking of: If you use on PageBundle, implement there (when this editable part has nothing to do with PageBundle in the first place).

> ... and I am using the 3.0 block version for many months without any problems which is a problem since a few days. Reverting a newly introduced BC break should be no general BC break.

Again, your usage might not be the general usage.

To me it looks like:

1. On 2.x we were relying on MenuBlockService from BlockBundle
2. MenuBlockService was marked as final (it was removed final by you on 4.x) (questionable choice)
3. You tried to not extend MenuBlockService on Seo3.x, but removing at the same time some features (also duplicating code between block bundle and seo bundle, questionable choice)
4. You also mentioned that breadcrumb shouldn't be a block
5. There are some issues talking about losing features of the breadcrumb block service that were closed without any solution

Giving this, to me the most reasonable choice was:

1. Partially revert the point 3., by implementing missing features from 2.x
2. For 4.0.0 if someone cares enough and provides clear upgrade path, solve this mess by not having breadcrumb as block in the first place.

Because if we want to move forward in a good pace, not needing several months of focused work on only one bundle, just to release a new major, we need to guarantee that deprecations, removal of code, and everything that might relate to another bundle is taken care with full knowledge on how that code is used across Sonata.

Otherwise we might end up in the same situation we are trying to avoid:

1. Not clear upgrade paths for deprecations (take a look at how hard was to solve the thing with SonataBlockBundle interfaces, or how hard is getting to upgrade SonataPageBundle, since multiple things were not taken into account)
2. Things changed without looking for usages or without taking care about other Sonata packages (I don't care about each project individually, I care about Sonata)
3. PRs merged fast, hard work later when you try to release anything, and have to solve issues you didn't create.

VincentLanglet:
Can you add the link of the issues you have in mind @jordisala1991 ?

VincentLanglet:
> You tried to not extend MenuBlockService on Seo3.x, but removing at the same time some features (also duplicating code between block bundle and seo bundle, questionable choice)

Currently, the MenuBlockService has a `menu_name` option
https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Block/Service/MenuBlockService.php#L105
which is used here:
https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Block/Service/MenuBlockService.php#L196

The previous method is overrided here: https://github.com/sonata-project/SonataSeoBundle/blob/3.x/src/Block/Breadcrumb/BaseBreadcrumbMenuBlockService.php#L110
which means that the `menu_name` option is useless.
But if it's an editable block, it means that you'll see a field to edit the `menu_name`, 
https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Block/Service/MenuBlockService.php#L155
and a validation
https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Block/Service/MenuBlockService.php#L87-L96

So we're wanting:
- Almost the same `configureSettings` method (except the menu_name field) ; and then we'll add more settings (menu_template, include_homepage_link, context)
- Almost the same `configureEditForm` method (except the menu_name field) ; but we need to add a way to configure the 3 extra settings (which is not possible now)
- Not the same `validate` method
- Almost the same `execute` method, except that we would like to override the `getMenu` method

VincentLanglet:
What about something like this:
https://github.com/sonata-project/SonataBlockBundle/pull/1112
https://github.com/sonata-project/SonataSeoBundle/pull/692
https://github.com/sonata-project/SonataPageBundle/pull/1612
@core23 @jordisala1991 ?

jordisala1991:
> Can you add the link of the issues you have in mind @jordisala1991 ?

Yes.

> MenuBlockService was marked as final (it was removed final by you on 4.x) (questionable choice)

https://github.com/sonata-project/SonataBlockBundle/pull/594
https://github.com/sonata-project/SonataSeoBundle/issues/443
https://github.com/sonata-project/SonataBlockBundle/pull/761

> You tried to not extend MenuBlockService on Seo3.x, but removing at the same time some features (also duplicating code between block bundle and seo bundle, questionable choice)

https://github.com/sonata-project/SonataSeoBundle/pull/399 (notice that there is no mention about losing features)

> There are some issues talking about losing features of the breadcrumb block service that were closed without any solution

https://github.com/sonata-project/SonataSeoBundle/commit/84355f6ed55d7a3c95dcfdf2bbfbc1f6ee385eee#r42221263
https://github.com/sonata-project/SonataSeoBundle/issues/446 (closed in favour of the next one)

> You also mentioned that breadcrumb shouldn't be a block

https://github.com/sonata-project/SonataSeoBundle/issues/520 (closed because no interest: 6months)

Also there was an initial attempt at doing another kind of implementation (similar to what is now in the code):

https://github.com/sonata-project/SonataSeoBundle/pull/356

Closed because of the lack of interest.

VincentLanglet:
> What about something like this: [sonata-project/SonataBlockBundle#1112](https://github.com/sonata-project/SonataBlockBundle/pull/1112) #692 [sonata-project/SonataPageBundle#1612](https://github.com/sonata-project/SonataPageBundle/pull/1612) @core23 @jordisala1991 ?

FYI: Three PRs are now green. This seems to be an OK solution for solving
- The bc breaks
- The SEO/Page issues

There is only one things to check before merging this:
The translations in
https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Resources/translations/SonataBlockBundle.en.xliff#L57-L76
are also here
https://github.com/sonata-project/SonataPageBundle/blob/4.x/src/Resources/translations/SonataPageBundle.en.xliff#L858-L885

IMHO, they should be removed from PageBundle, in order to rely on blockBundle or Seo ones.

jordisala1991:
To me that solution looks indeed better than this one. I can take a look at translations but not today I think.

core23:
Forked this project, feel free to fix the design issue with an other PR

PR #696, opened 2022-10-06
Change menu label to Breadcrumb menu

eerison:
## Adding a name more specific for `breadcrumb menu`

I am targeting this branch, because This issue is into this branch.

## Changelog

```markdown
### Changed
- Changed `menu label` to `Breadcrumb menu`
```

It will sove the issue when we depends the block bundle and seo bundle, like in page bundle

Screenshot 2022-10-06 at 10 35 11

### DIFF
--- a/src/Resources/translations/SonataSeoBundle.de.xliff
+++ b/src/Resources/translations/SonataSeoBundle.de.xliff
@@ -14,6 +14,10 @@
         <source>sonata.seo.block.email.share_button</source>
         <target>E-Mail teilen</target>
       </trans-unit>
+      <trans-unit id="sonata_seo_block_menu_breadcrumb">
+        <source>sonata.seo.block.menu.breadcrumb</source>
+        <target>Breadcrumb Menü</target>
+      </trans-unit>
       <trans-unit id="sonata_seo_block_facebook_like_box">
         <source>sonata.seo.block.facebook.like_box</source>
         <target>Facebook - Likebox</target>
--- a/src/Block/Breadcrumb/BaseBreadcrumbMenuBlockService.php
+++ b/src/Block/Breadcrumb/BaseBreadcrumbMenuBlockService.php
@@ -55,8 +55,8 @@ abstract class BaseBreadcrumbMenuBlockService extends AbstractMenuBlockService i
      */
     public function getMetadata(): MetadataInterface
     {
-        return new Metadata('sonata.block.service.menu', null, null, 'SonataBlockBundle', [
-            'class' => 'fa fa-bars',
+        return new Metadata('sonata.seo.block.menu.breadcrumb', null, null, 'SonataSeoBundle', [
+            'class' => 'fa fa-angle-double-right',
         ]);
     }
 
--- a/src/Resources/translations/SonataSeoBundle.en.xliff
+++ b/src/Resources/translations/SonataSeoBundle.en.xliff
@@ -14,6 +14,10 @@
         <source>sonata.seo.block.email.share_button</source>
         <target>Share via E-Mail</target>
       </trans-unit>
+      <trans-unit id="sonata_seo_block_menu_breadcrumb">
+        <source>sonata.seo.block.menu.breadcrumb</source>
+        <target>Breadcrumb Menu</target>
+      </trans-unit>
       <trans-unit id="sonata_seo_block_facebook_like_box">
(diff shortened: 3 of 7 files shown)


eerison:
@VincentLanglet Can I let the others translation missing it?
how do you usually handle new translations for all languages?

VincentLanglet:
> @VincentLanglet Can I let the others translation missing it? how do you usually handle new translations for all languages?

I generally try to add translation thanks to some knowledge, review or translator.

In french, `breadcrumb menu` is `Menu fil d'Ariane`

eerison:
> > @VincentLanglet Can I let the others translation missing it? how do you usually handle new translations for all languages?
> 
> I generally try to add translation thanks to some knowledge, review or translator.
> 
> In french, `breadcrumb menu` is `Menu fil d'Ariane`

Ok I asked some guys and They suggested to translate `Menu` only, because `Breadcrumb` doesn't make sense!

eerison:
Just to have a different icon, what do you think about this one

https://fontawesome.com/v5/icons/angle-double-right?s=solid&f=classic

or should I keep the same (bars)?

VincentLanglet:
> Just to have a different icon, what do you think about this one
> 
> https://fontawesome.com/v5/icons/angle-double-right?s=solid&f=classic
> 
> or should I keep the same (bars)?

This icon might be a good idea indeed

eerison:
Screenshot 2022-10-06 at 11 17 30
Ok then it stays like this, I liked it 💅🏼

VincentLanglet:
We're changing from `fa` to `fas` are we sure font awesome 5 is used by all the seoBundle users ?
Or should we stay with something like `<i class="fa fa-angle-double-right"></i>` ?

eerison:
> We're changing from `fa` to `fas` are we sure font awesome 5 is used by all the seoBundle users ? Or should we stay with something like `<i class="fa fa-angle-double-right"></i>` ?

hmmm well I was following the sonata admin 4!

eerison:
> > We're changing from `fa` to `fas` are we sure font awesome 5 is used by all the seoBundle users ? Or should we stay with something like `<i class="fa fa-angle-double-right"></i>` ?
> 
> hmmm well I was following the sonata admin 4!

I tested and it works on both way `fa` or `fas`

eerison:
But as we require [block bundle 4](https://github.com/sonata-project/SonataSeoBundle/blob/3.x/composer.json#L25), I guess the user is more or less forced to use sonata 4, don't you?

VincentLanglet:
> But as we require [block bundle 4](https://github.com/sonata-project/SonataSeoBundle/blob/3.x/composer.json#L25), I guess the user is more or less forced to use sonata 4, don't you?

We dont force to use sonataAdmin 4. People might use just BlockBundle and SeoBubdle maybe...

I dunno where come from this icon, so I dunno if we can use font awesome 5 without a BC break...

eerison:
well Ok lets use `fa`

VincentLanglet (approved):

src/Resources/translations/SonataSeoBundle.de.xliff
```diff
@@ -14,6 +14,10 @@
         <source>sonata.seo.block.email.share_button</source>
         <target>E-Mail teilen</target>
       </trans-unit>
+      <trans-unit id="sonata_seo_block_menu_breadcrumb">
+        <source>sonata.seo.block.menu.breadcrumb</source>
+        <target>Breadcrumb Menü</target>
```
VincentLanglet, line 19: Where did you find your translations ?
I would say it's weird Breadcrumb is said Breadcumb in De, It, Nl and Ru

VincentLanglet (changes requested):

src/Resources/translations/SonataSeoBundle.de.xliff
eerison: I asked some german opinion internally, And was suggested translate just `Menu`

I got two option `Breadcrumb-Menü` or `Sekundäre Navigation`

eerison: and I took the first one

VincentLanglet (approved):

eerison:
@jordisala1991 could you review please 👀

jordisala1991 (approved):

jordisala1991:
Thank you @eerison

PR #697, opened 2022-10-15
Remove Old translations

eerison:
## Remove Old translations

I am targeting this branch, because the translation related with the old cold does not exist in current code anymore.
This translation was introduced in this commit: https://github.com/sonata-project/SonataSeoBundle/commit/215e703fb97dc5017e8383d424cc90e43cf656bc#diff-ae5aa53fecd1379bdf277729ce8275f3d2f167088dffe46c27dbdc2aa652bb0d

## Changelog

```markdown

### Removed
- Removed old translation related with `social blocks` (`twitter`, `facebook` , `email` and so on)
```


VincentLanglet:
I think we're forced to target 4.x because this is still a BC break @eerison.
We don't use those translations anymore, but some user could...

eerison:
> I think we're forced to target 4.x because this is still a BC break @eerison. We don't use those translations anymore, but some user could...

Make sense! should I change just the target and solve the conflicts or should I add some UPGRADE notes?

VincentLanglet:
No need for upgrade note IMHO

eerison:
Done!

VincentLanglet (dismissed):

jordisala1991 (dismissed):

VincentLanglet:
Can you fix the conflicts @eerison then I'll merge it

VincentLanglet (approved):

PR #701, opened 2022-12-08
DevKit updates for 3.x branch

SonataCI:
## Changelog

```markdown
### Deprecated
- SeoAwareTrait
```


VincentLanglet:
Phpstan reports that the SeoAwareTrait is not used (and not analyzed).

We never provided such trait for others service, I don't see a reason to have one here.
- I can't say it works because there is no tests about it and the `@required` is on the property rather than the setter.
- People could just use classic dependency injection.

